PR #1, opened 2023-08-30
Move to .config directory

mattgauntseo-sentry:
Added some comment and questions after reading through the script

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 
-root = os.path.expanduser("~/.sentry-dev")
+root = os.path.expanduser("~/.config/sentry-dev")
 src_root = f"{root}/devenv"
 pythons_root = f"{root}/pythons"
 venv_root = f"{root}/virtualenvs"
--- a/install-devenv.sh
+++ b/install-devenv.sh
@@ -30,6 +30,7 @@ esac
 
 archive="cpython-3.11.4+20230726-${platform}-apple-darwin-install_only.tar.gz"
 
+# Question: What make a temp directory and then rm -rf it?
 tmpd=$(mktemp -d)
 trap 'rm -rf "$tmpd"' EXIT
 
--- a/install-devenv.sh
+++ b/install-devenv.sh
@@ -7,15 +7,17 @@ fi
 
 # TODO: idempotent
 
-devenv_root="${HOME}/.sentry-dev"
+devenv_root="${HOME}/.config/sentry-dev"
 devenv_bin="${devenv_root}/bin"
 devenv_python_root="${devenv_root}/python"
 
 mkdir -p "$devenv_python_root" "$devenv_bin"
 
+# Default to intel chips
 platform=x86_64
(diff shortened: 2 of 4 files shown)


constants.py
```diff
@@ -2,7 +2,7 @@
 
 import os
 
-root = os.path.expanduser("~/.sentry-dev")
+root = os.path.expanduser("~/.config/sentry-dev")
```
joshuarli, line 5: i ultimately decided against this because this is the root where we put everything (pythons, virtualenvs, etc.), and ~/.config should be strictly for configuration files only

install-devenv.sh
```diff
@@ -25,6 +30,7 @@ esac
 
 archive="cpython-3.11.4+20230726-${platform}-apple-darwin-install_only.tar.gz"
 
+# Question: What make a temp directory and then rm -rf it?
```
joshuarli, line 33: Don't quite understand this question? You've described what happens in the following 2 lines.

install-devenv.sh
```diff
@@ -36,13 +42,17 @@ curl -fsSL \
 
 echo "${sha256}  ${tmpd}/${archive}" | /usr/bin/shasum -a 256 --check --status
 
+# Question: What is the --strip-components doing here?
```
joshuarli, line 45: it conveniently removes a top level directory that is present in the archive so i don't have to do an extra `mv`

install-devenv.sh
```diff
@@ -36,13 +42,17 @@ curl -fsSL \
...
 tar --strip-components=1 -C "$devenv_python_root" -x -f "${tmpd}/${archive}"
 
+# Question: Is there a reason to use the SSH URL here?
```
joshuarli, line 48: yes, by making sure we clone over SSH we exercise that engineers have their github ssh keys set up correctly

i'll be coming up with a separate github ssh key setup thing soon

install-devenv.sh
```diff
@@ -36,13 +42,17 @@ curl -fsSL \
...
 uri='git@github.com:getsentry/devenv'
 [[ $CI ]] && uri='https://github.com/getsentry/devenv.git'
 git -C "$devenv_root" clone -q --depth=1 "$uri"
 
+# Question: What does this syntax mean / do?
```
joshuarli, line 53: uhhhhhh actually this is wrong and i need to fix it - it should be a substring test

install-devenv.sh
```diff
@@ -36,13 +42,17 @@ curl -fsSL \
...
 if [[ ":$PATH:" != *":$devenv_bin:"* ]]; then
+    # Question: Should this avoid making a .zshrc file if it doesnt exist?
```
joshuarli, line 55: no, we should make sure it exists

install-devenv.sh
joshuarli: it's used as a temporary place to download the archive to, then checksum check it, before finally unpacking it into the desired place

now that i'm thinking about it more, it would be better to download to `~/.cache/sentry-dev` actually

PR #3, opened 2023-09-07
feat: doctor example with brew packages check, autofix, recheck

joshuarli:
I dunno, using python classes for this feels kind of clunky? Read below, and also I have no idea how to type the import discovery stuff - maybe a sign that it's just a bad idea?

An alternative: just executables? Look for *.sh, *.fix.sh, *.py, *.fix.py. If people want to have the venv activated then it's just sourcing for sh and the venv's python3 in the shebang for py. I think this is simpler for us and the contributors? On our end we'd just need to add devenv to python path.
- edit: Ah virtualenv was pretty easy, look here: https://github.com/getsentry/sentry/commit/925cebba3022e068a97d5e9eb6a2401f567c38fb#diff-e453e0328e00ec4ab312a1cd274393f345b342ffbf49a0d41f6d507c79fd6aa5

...

This is the basic idea for a `doctor` command that looks in `REPO/devenv/checks/*.py` and for each module it finds, it imports `module.Check` which we define an ABC for. These checks are able to import from `devenv`.

Example check: https://github.com/getsentry/sentry/pull/55486/files#diff-7b413695cd76c079baa4c828edae606868c8f855fc2b38bac787b290156c1159

Checks are added conditionally based on tags, then executed in parallel. For each check we run `check` first, then failing that we run `fix`, and then a final `check`.

image


doctor.py
```diff
@@ -2,19 +2,96 @@
 
 import argparse
 from collections.abc import Sequence
+from concurrent.futures import ThreadPoolExecutor
```
bukzor, line 5: The documentation for both  [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) and [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor) note that deadlocks can occur. In practice, since this is an API we'll be distributing to a wide audience with varying experience levels, it means deadlocks *will* occur.

Our fixer API needs to be more foolproof than that, IMO.

> Deadlocks can occur ...

Which in practice means deadlocks *will* occur.

doctor.py
```diff
@@ -2,19 +2,96 @@
...
+from pkgutil import walk_packages
 
-help = "Diagnose common issues."
+help = "Diagnose and attempt to fix common issues."
```
bukzor, line 8: Make a separate `fix` subcommand (which shares much of its code with the `diagnose` subcommand). There are more than a few times where readonly diagnosis is much more appropriate.

doctor.py
```diff
@@ -2,19 +2,96 @@
...
+        ok, msg = result
+        if ok:
+            print(f"✅ check {check}")
+            continue
+        print(f"❌ check {check}{msg}")
+        all_ok = False
+
+    if all_ok:
+        print("\nLooks good to me now!")
+        return 0
```
bukzor, line 94: This function is 85 lines long, indicating it should be broken into 4+ smaller functions.
This is good for testability but also for high-level refactorizations and reusability.

lib.py
```diff
@@ -1,12 +1,35 @@
...
+import os
+from abc import ABC
+from abc import abstractmethod
 from functools import cache
+from subprocess import CalledProcessError
+from subprocess import run
+
+
+class MetaCheck(ABC):
+    tags: set
```
bukzor, line 12: What is? There's no doc, here.

lib.py
```diff
@@ -1,12 +1,35 @@
 from __future__ import annotations
```
bukzor, line 1: It's okay to have a directory named "lib" but a `lib.py` file is equivalent to a `stuff.py` -- it's too ambiguous what should go in there and becomes a 9000LOC trash heap too easily.

* setup.py / pypackage.toml
* `src/`
    * `$package_name` -- TBD? perhaps: getsentry_devenv, or if it's meant to help selfhosted too: sentry_devenv
       * `doctor.py`
       * `fixer_lib` -- if we want polyglot support (imo yes), expose each of these via `devenv fixer-lib xyz` subcommands
           * `brew_packages.py`
           * `gitroot.py`

lib.py
IanWoodard: It is meant to allow us to mark checks with specific tags so we can, for instance, not run multiple checks that try to edit the same file at once, or being able to only run checks for a specific tag. With that said, we should add some documentation around that.

doctor.py
IanWoodard: Agreed. This can be something I can address in the follow up.

lib.py
joshuarli: I'll divide it into `lib/*.py`.

As for setup.py and package_name - I made it so that devenv isn't installed via pip, it just gets called as a module directly. See the `devenv` executable.

doctor.py
joshuarli: I don't think this is a concern for the TPE we're using - check authors are not gonna be able to reference other futures

joshuarli:
i've:
- added a `--check-only`
- written an example check to execute python in the sentry virtualenv: https://github.com/getsentry/sentry/pull/55486/files
  - that PR still needs a little more work in the .envrc, and i need to finish getting `devenv sync` up to par
- broke up `lib.py`
- added some CI to exercise `devenv doctor`

i think this should be good to merge and Ian can work on followups like:
- breaking `doctor.py` down into smaller functions
- writing checks for sentry devservices

some thoughts
- checks are mostly network or i/o bound so I think it would be fine to keep it simpler with one thread per job - if we provided a --jobs i don't really see a need for it but have no strong opinion if someone implements that

bukzor (approved):

PR #4, opened 2023-09-18
feat: bootstrap skeleton

joshuarli:
skeleton for bootstrap, including interactively configuring coderoot.

Will be following up with individual portions of bootstrap after; trying to break this up into smaller reviewable PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,12 @@ jobs:
 
       - name: bootstrap
         run: |
-          true
-          # TODO need to provide input for configuring coderoot
+          cat <<EOF >  ~/.config/sentry-devenv/config.ini
+[devenv]
+coderoot = /Users/josh/dev
+EOF
+          devenv bootstrap
+          false
 
       - name: test
         shell: zsh -i {0}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,7 @@ jobs:
           ./install-devenv.sh "${{ github.event.pull_request && github.head_ref || github.ref_name }}"
 
       - name: bootstrap
+        shell: zsh -i {0}
         run: |
           mkdir -p ~/.config/sentry-devenv
           cat <<EOF > ~/.config/sentry-devenv/config.ini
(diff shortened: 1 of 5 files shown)


main.py
```diff
@@ -71,18 +76,36 @@ def main(argv: Sequence[str] | None = None) -> int:
...
+    coderoot = config["devenv"]["coderoot"]
+
+    if args.command == "bootstrap":
+        return bootstrap.main(coderoot, remainder)
+
+    # the remaining tools are repo-specific
     if not args.nocoderoot and not args.pwd.startswith(os.path.expanduser(f"~/{coderoot}")):
         print(
-            f"You aren't in your code root (~/{coderoot})!"
+            f"You aren't in your code root (~/{coderoot})! "
```
IanWoodard, line 108: Is there a reason for the extra space here?

main.py
```diff
@@ -71,18 +76,36 @@ def main(argv: Sequence[str] | None = None) -> int:
...
+
+        os.makedirs(config_root, exist_ok=True)
+        with open(config_path, "w") as f:
+            config.write(f)
+        print(f"If you made a mistake, you can edit {config_path}.")
 
-    # TODO: read a well-known json for preferences
-    coderoot = "dev"
+    config.read(config_path)
+    coderoot = config["devenv"]["coderoot"]
```
IanWoodard, line 100: We might want to check that the coderoot actually exists

IanWoodard (commented):
Looking good so far

main.py
joshuarli: Yep, it's concatenated with the string below.

main.py
joshuarli: It doesn't have to exist for the bootstrap step (as bootstrap will be creating the coderoot if it doesn't exist)
and the subsequent `args.pwd.startswith(coderoot)` will fail if coderoot doesn't exist

IanWoodard (approved):

.github/workflows/test.yml
```diff
@@ -24,10 +24,19 @@ jobs:
           # should be able to be run from anywhere
           ./install-devenv.sh "${{ github.event.pull_request && github.head_ref || github.ref_name }}"
 
+      - name: bootstrap
+        shell: zsh -i {0}
+        run: |
+          mkdir -p ~/.config/sentry-devenv
+          cat <<EOF > ~/.config/sentry-devenv/config.ini
+          [devenv]
+          coderoot = /Users/josh/dev
```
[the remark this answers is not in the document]
joshuarli: whoops - i'll change this to $HOME/dev

PR #5, opened 2023-09-18
feat: automate github setup

joshuarli:
This automates setting up an ssh keypair for github so that people can clone private repos, and then clones sentry+getsentry into coderoot.


install-devenv.sh
```diff
@@ -38,12 +38,7 @@ echo "${sha256}  ${devenv_cache}/${archive}" | /usr/bin/shasum -a 256 --check --
 tar --strip-components=1 -C "$devenv_python_root" -x -f "${devenv_cache}/${archive}"
 
 if ! [[ -d "${devenv_root}/devenv" ]]; then
-    if [[ $CI ]]; then
-        git -C "$devenv_root" clone -q -b "$branch" --depth=1 'https://github.com/getsentry/devenv.git'
-    else
-        # This makes sure we clone over ssh so that github keys are working.
-        git -C "$devenv_root" clone -q -b "$branch" --depth=1 'git@github.com:getsentry/devenv'
-    fi
+    git -C "$devenv_root" clone -q -b "$branch" --depth=1 'https://github.com/getsentry/devenv.git'
```
joshuarli, line 41: Cloning over https will work at this point as it's unauthenticated. When I wrote the previous code, `devenv` was a private repo at the time.

mattgauntseo-sentry (approved):

bootstrap.py
```diff
@@ -32,4 +29,82 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
         return 1
 
+    # xcode-select --install will take a while,
+    # and involves elevated permissions with a GUI,
+    # so best to just let the user go through that separately then retrying,
+    # rather than waiting for it.
+    # There is a way to perform a headless install but it's more complex
+    # (refer to how homebrew does it).
+    try:
+        xcode_git = proc.run(("/usr/bin/xcrun", "-f", "git"))
```
mattgauntseo-sentry, line 39: Can you break this up into functions to avoid one large main block

mattgauntseo-sentry (changes requested):

bootstrap.py
joshuarli: This is pretty much as good as it gets... it looks largeish but it's mainly that big block of text and a few long subprocess argument lists that get spread out. It's really not that large in terms of cyclomatic complexity

mattgauntseo-sentry, line 39: Nit: Would it make sense to change `xcode_git` to `git_binary` because it looks like it doesn't matter that its xcode git, it just matters that it's a git_binary

bootstrap.py
```diff
@@ -32,4 +29,82 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
+    except RuntimeError:
+        print("Run xcode-select --install, then come back to bootstrap when done.")
```
mattgauntseo-sentry, line 41:
```suggestion
        print("Failed to find git. Run xcode-select --install, then re-run bootstrap when done.")
```

lib/github.py
```diff
@@ -0,0 +1,71 @@
...
+from devenv.lib import fs
+from devenv.lib import proc
+
+
+def check_ssh_access(git: str) -> bool:
+    if CI:
+        return True
+    try:
+        # The remote returns code 1 when successfully authenticated.
+        proc.run(("ssh", "-T", "git@github.com"))
```
mattgauntseo-sentry, line 17: Wouldn't it be cleaner to get the result and check the code for `1` instead of catching an error and checking for a string in the output?

bootstrap.py
```diff
@@ -32,4 +29,82 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
+                if CI
+                else ("git@github.com:getsentry/sentry",)
+            )
+            proc.run_stream_output(
+                (
+                    xcode_git,
+                    "-C",
+                    coderoot,
+                    "clone",
+                    # https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/
```
mattgauntseo-sentry, line 82:
```suggestion
                    # Download all reachable commits and trees while fetching blobs on-demand
                    # https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/
```

bootstrap.py
```diff
@@ -32,4 +29,82 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
+                    "--filter=blob:none",
+                    *additional_flags,
+                ),
+                exit=True,
+            )
+        if not CI and not os.path.exists(f"{coderoot}/getsentry"):
```
mattgauntseo-sentry, line 88: Should this be indented less? 

At the moment it looks like running this with `--repo="sentry"` it'll clone both sentry and getsentry repo, but if you pass in getsentry nothing will be cloned - but I might be missing something / reading GitHub wrong.

bootstrap.py
```diff
@@ -32,4 +29,82 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
+                    "clone",
+                    "--filter=blob:none",
+                    "git@github.com:getsentry/getsentry",
+                ),
+                exit=True,
+            )
+
+    # TODO: install brew
+
+    if args.repo == "sentry":
```
mattgauntseo-sentry, line 103: Since this is the second args.repo == "sentry" check, does it make sense to have something like:

```
 # Setup environment
check_github_access()

# Get repos
if args.repo == "sentry":
    sentry_path = checkout_sentry()
    brew_install(sentry_path)
if args.repo == "getsentry":
    getsentry_path = checkout_getsentry()
    brew_install(getsentry_path)
 
install_volta()
install_direnv()
....
```

lib/github.py
mattgauntseo-sentry: Is it worth returning a message along with False? It seems like the two tests are checking different things and the context is helpful to the user (either the SSH key isn't added to their profile correctly or they haven't setup SSO correctly.

lib/proc.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Tuple
+from typing import Union
+
+
+def run(
+    cmd: Tuple[str, ...], exit: bool = False, **kwargs: Union[str, bool, dict[str, str]]
+) -> str:
+    kwargs["check"] = True
+    kwargs["capture_output"] = True
+    try:
```
mattgauntseo-sentry, line 14: Could `run()` call `run_stream_output()` with the extra kwargs if `run_stream_output()` returns the result of its subprocess.run() call?

Would remove some of the duplication.

mattgauntseo-sentry (approved):

bootstrap.py
joshuarli: you missed this which is towards the beginning

```
    if args.repo == "getsentry":
        # Setting up sentry means we're setting up both repos.
        args.repo = "sentry"
```

bootstrap.py
joshuarli: ah this is an artifact, i intend to remove it

and no, bootstrapping sentry+getsentry is one and the same

lib/proc.py
joshuarli: i'm not sure how to deduplicate this in a nice way - if i could have i would have - run_stream_output would have to reraise the CalledProcessError if capture_output=True and run would still need to print more detail so overall not much is being deduplicated really

bootstrap.py
joshuarli: thinking about this a bit more i'm just gonna use the git on PATH after xcode git is installed (/usr/bin/git will work afterwards)

lib/github.py
joshuarli: no, there are probably other scenarios where github returns 1 and has a different error message - we're specifically interested in this one

joshuarli: added a print in the case that ssh key isn't added correctly
and SSO not working implies the ssh key is setup correctly

PR #6, opened 2023-09-26
feat(bootstrap): brew

joshuarli:
This installs brew and sentry's brew dependencies.


lib/brew.py
```diff
@@ -0,0 +1,52 @@
...
+
+from devenv.lib import fs
+from devenv.lib import proc
+
+repo_path = "/opt/homebrew"
+
+
+def install() -> None:
+    if which("brew") == f"{repo_path}/bin/brew":
+        return
```
mattgauntseo-sentry, line 14: Can you add a comment about this check.

I.e. does this mean than brew bundle has been run before?

lib/brew.py
```diff
@@ -0,0 +1,52 @@
...
+
+    shellrc = fs.shellrc()
+    print("You may be prompted for your password to install homebrew.")
+    while True:
```
mattgauntseo-sentry, line 18: Should we limit how many times this runs?

lib/brew.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
+
+import os
+from shutil import which
+
+from devenv.lib import fs
+from devenv.lib import proc
+
+repo_path = "/opt/homebrew"
```
mattgauntseo-sentry, line 9: Can you rename this to something like "brew_install_path". I started to think we were installing brew into the sentry repo

mattgauntseo-sentry (approved):

lib/brew.py
joshuarli: done. no, it's just concerned with installing brew

lib/brew.py
```diff
@@ -0,0 +1,52 @@
...
+""",
+                ),
+                exit=False,
+            )
+        except RuntimeError:
+            continue
+        break
+
+    if not os.path.exists(f"{repo_path}/bin/brew"):
+        os.makedirs(repo_path, exist_ok=True)
```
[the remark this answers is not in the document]
joshuarli: ah you're right, sorry this is an artifact before i added the which("brew") in the top

lib/brew.py
joshuarli: No, people can ^C out at any time.

PR #7, opened 2023-09-26
feat(bootstrap): volta

joshuarli:
This installs volta without curlpipebash'ing the volta installer script.


pythons.py
joshuarli, line 1: figured i would also reuse the new lib/archive.py for the runtime python

lib/volta.py
```diff
@@ -0,0 +1,46 @@
...
+from shutil import which
+
+from devenv.constants import root
+from devenv.lib import archive
+from devenv.lib import fs
+from devenv.lib import proc
+
+VOLTA_HOME = f"{root}/volta"
+
+_version = "1.1.1"
```
bukzor-sentryio, line 14: Why such an ancient version?
The current version is 1.18.1:
https://github.com/Homebrew/homebrew-cask/blob/f398e64fd/Casks/v/volta.rb#L4

Their history doesn't go as far back as 1.1.1, but 1.15.0 is four months old:
https://github.com/Homebrew/homebrew-cask/blame/350bb474/Casks/volta.rb#L4C6-L4C19

lib/volta.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if which("volta") == f"{unpack_into}/volta":
+        return
+
+    suffix = "-aarch64" if platform.machine() == "arm64" else ""
+    name = f"volta-{_version}-macos{suffix}.tar.gz"
+    url = "https://github.com/volta-cli/volta/releases/download/" f"v{_version}/{name}"
+
+    archive_file = archive.download(url, _sha256[name])
+    archive.unpack(archive_file, unpack_into)
```
bukzor-sentryio, line 32: Isn't `subprocess.check_call('brew install --cask volta')` equivalent (but better maintained)?
Note the similarities (and differences):
https://github.com/Homebrew/homebrew-cask/blob/master/Casks/v/volta.rb

bukzor-sentryio (changes requested):

lib/volta.py
joshuarli: that cask only began existence this May so i'm wary of maintainership https://github.com/Homebrew/homebrew-cask/commits/aa461148bbb5119af26b82cccf5003e2b4e50d95/Casks/volta.rb?browsing_rename_history=true&new_path=Casks/v/volta.rb&original_branch=master

i would prefer to have control over how it's installed and it's pretty easy to do (as you can see)

lib/volta.py
joshuarli: oh cool, i was just going with whatever sentry-bootstrap already had, will upgrade

joshuarli: No idea what that 1.18.1 is, 1.1.1 is the latest upstream: https://github.com/volta-cli/volta

IanWoodard (approved):

mattgauntseo-sentry (approved):

PR #9, opened 2023-09-26
feat(bootstrap): complete sentry virtualenv

joshuarli:
This creates a sentry+getsentry virtualenv (via `devenv sync`), and calls `make bootstrap` to finish setting sentry up. This completes feature parity with sentry-bootstrap and at this point we can start onboarding with this, but I'd first like to f/u with:
- testing on intel laptop
- making it seamless for current engineers to use new devenv tooling
- adding sentry-sdk so we can track devenv errors
- get sentry devenv-compat branch finished and merged
  - devenv sync needs to do the js and migrations stuff

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,8 @@ jobs:
           #       macos GitHub runners
           devenv bootstrap
 
+          # TODO: remove this when it's merged
           git -C $HOME/dev/sentry checkout origin/devenv-compat
+
           cd $HOME/dev/sentry
           direnv allow
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,11 +39,7 @@ jobs:
         env:
           SHELL: zsh
         run: |
-          {
-              devenv --nocoderoot doctor --check-only
-              [[ $? = 1 ]]
-          }
-          touch foo
+          devenv --nocoderoot doctor --check-only || touch foo
           devenv --nocoderoot doctor --check-only
           rm foo
           devenv --nocoderoot doctor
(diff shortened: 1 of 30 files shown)


.github/workflows/test.yml
```diff
@@ -18,30 +18,28 @@ jobs:
       - uses: pre-commit/action@v3.0.0
 
       - name: install
+        shell: zsh -i -euo pipefail {0}
```
asottile-sentry, line 21: zsh isn't really a scripting language -- we should still be using bash

bootstrap.py
```diff
@@ -104,9 +107,41 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
+
+        # make bootstrap should be ported over to devenv sync,
+        # as it applies new migrations as well and so would need to ensure
+        # the appropriate devservices are running
+        proc.run_stream_output(
+            (
+                shell,
+                "-i",
+                "-euo",
+                "pipefail",
```
asottile-sentry, line 125: we shouldn't need to run a shell  `venv/bin/activate` just sets `PATH=venv/bin:$PATH`

sync.py
```diff
@@ -46,10 +46,10 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
     if repo == "getsentry":
         repo = "sentry"
         reporoot = f"{reporoot}/../sentry"
-        os.chdir(reporoot)
 
-    reporoot = context["reporoot"]
+    os.chdir(reporoot)
```
asottile-sentry, line 50: chdir is usually a smell in library code since cwd is global

.github/workflows/test.yml
joshuarli: people are gonna be using zsh on macos though, but i guess we can switch since in the future i'd like this to work for people on linux as well

asottile-sentry (approved):

PR #10, opened 2023-09-27
Adding the ability to skip fixes

IanWoodard:
By allowing users to opt in to fixes, it gives them more control of their developer environment and allows them to skip potentially buggy fixes without having to be blocked. It also ensures transparency and allows users to see what changes are going to be made so they can review them before they are applied.

### DIFF
--- a/doctor.py
+++ b/doctor.py
@@ -67,25 +67,29 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
         if args.check_only:
             return 1
 
-    print("\nAttempting to run fixes for the checks that failed...")
-    futures = {}
-    results = {}
+    print("\nThe following checks need to be fixed:")
+    skip = []
     for check in retry:
-        futures[check] = executor.submit(check.fix)
-    for check, future in futures.items():
-        results[check] = future.result()
-
-    for check, result in results.items():
-        ok, msg = result
-        if ok:
-            print(f"✅ fix: {check.name}")
-            continue
-        print(f"❌ fix: {check.name}{msg}")
+        print(f"❌ {check.name}")
+        if input(f"Do you want to attempt to fix {check.name}? (y/n): ").lower() == "y":
+            future = executor.submit(check.fix)
+            result = future.result()
+            ok, msg = result
+            if ok:
+                print(f"✅ fix: {check.name}")
+            else:
+                print(f"❌ fix: {check.name}{msg}")
+        else:
+            print(f"Skipping {check.name}.")
+            skip.append(check)
 
     print("\nChecking again...")
     futures = {}
     results = {}
     for check in retry:
+        if check in skip:
+            print(f"⏭️ Skipping {check.name}.")
+            continue
         futures[check] = executor.submit(check.check)
     for check, future in futures.items():
         results[check] = future.result()
@@ -96,9 +100,9 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
     for check, result in results.items():
         ok, msg = result
         if ok:
-            print(f"✅ check {check.name}")
+            print(f"✅ check: {check.name}")
             continue
-        print(f"❌ check {check.name}{msg}")
+        print(f"❌ check: {check.name}{msg}")
         all_ok = False
 
     if all_ok:


.github/workflows/test.yml
```diff
@@ -44,5 +44,10 @@ jobs:
           touch foo
           devenv --nocoderoot doctor --check-only
           rm foo
-          devenv --nocoderoot doctor
+          expect <<EOF
+            spawn devenv --nocoderoot doctor
+            expect "Do you want to attempt to fix foo? (y/n): "
+            send "y\r"
```
mattgauntseo-sentry, line 50: I'm struggling to understand what this is doing and why send `y\r`, could you add a comment explaining what this is checking / doing?

doctor.py
```diff
@@ -67,25 +67,29 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
         if args.check_only:
             return 1
 
-    print("\nAttempting to run fixes for the checks that failed...")
-    futures = {}
-    results = {}
+    print("\nThe following checks need to be fixed:")
```
mattgauntseo-sentry, line 70: Should this statement be changed to:

`The following problems need to be fixed:` or `The following problems have been identified:`

doctor.py
```diff
@@ -67,25 +67,29 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
...
-        results[check] = future.result()
-
-    for check, result in results.items():
-        ok, msg = result
-        if ok:
-            print(f"✅ fix: {check.name}")
-            continue
-        print(f"❌ fix: {check.name}{msg}")
+        print(f"❌ {check.name}")
+        if input(f"Do you want to attempt to fix {check.name}? (y/n): ").lower() == "y":
```
mattgauntseo-sentry, line 74: Optional: Can we add a default value of y and change the input to `(Y/n)` ?

doctor.py
```diff
@@ -67,25 +67,29 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
...
+            future = executor.submit(check.fix)
+            result = future.result()
+            ok, msg = result
+            if ok:
+                print(f"✅ fix: {check.name}")
+            else:
+                print(f"❌ fix: {check.name}{msg}")
+        else:
+            print(f"Skipping {check.name}.")
```
mattgauntseo-sentry, line 83:
```suggestion
            print(f"⏭️ Skipping {check.name}.")
```

doctor.py
```diff
@@ -67,25 +67,29 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
...
+            skip.append(check)
 
     print("\nChecking again...")
     futures = {}
     results = {}
     for check in retry:
+        if check in skip:
+            print(f"⏭️ Skipping {check.name}.")
```
mattgauntseo-sentry, line 91:
```suggestion
            print(f"⏭️ Skipped {check.name}.")
```

mattgauntseo-sentry (approved):

.github/workflows/test.yml
IanWoodard: It is responding with "y" indicating that the fix should be performed.

doctor.py
IanWoodard: sure!

doctor.py
```diff
@@ -67,25 +67,33 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
         if args.check_only:
             return 1
 
-    print("\nAttempting to run fixes for the checks that failed...")
-    futures = {}
-    results = {}
+    print("\nThe following problems have been identified:")
```
joshuarli, line 70: This won't print everything upfront which is confusing. For example:

```
Running checks: foo, foo2
❌ check: foo/Users/josh/.local/share/sentry-devenv/devenv/foo doesn't exist
❌ check: foo2/Users/josh/.local/share/sentry-devenv/devenv/foo2 doesn't exist

The following problems have been identified:
❌ foo
Do you want to attempt to fix foo? (Y/n): n
⏭️ Skipping foo.
❌ foo2
Do you want to attempt to fix foo2? (Y/n): n
⏭️ Skipping foo2.

Checking again...
⏭️ Skipped foo.
⏭️ Skipped foo2.

Looks good to me now!
```

IanWoodard: Good catch

mattgauntseo-sentry: @joshuarli what do you find confusing about this? It seems ok to me.

Side note/future change: It would be nice to indent some of these logs to add some logical grouping.

```
Running checks: foo, foo2
    ❌ check: foo/Users/josh/.local/share/sentry-devenv/devenv/foo doesn't exist
    ❌ check: foo2/Users/josh/.local/share/sentry-devenv/devenv/foo2 doesn't exist


The following problems have been identified:
    ❌ foo
        Do you want to attempt to fix foo? (Y/n): n
        ⏭️ Skipping foo.

    ❌ foo2
        Do you want to attempt to fix foo2? (Y/n): n
        ⏭️ Skipping foo2.


Checking fixes worked as expected...
    ⏭️ Skipped foo.
    ⏭️ Skipped foo2.

Looks good to me now!
```

PR #11, opened 2023-09-29
Cleaning up the doctor

IanWoodard:
The goal of this PR is to break logic (especially duplicate logic) into their own functions, making the code more readable in the process. Additionally, I worked to improve the typing to add more safety. One thing I plan to add soon is the ability to not include a fix since not all checks will have fixes.

### DIFF
--- a/doctor.py
+++ b/doctor.py
@@ -5,6 +5,7 @@ from collections.abc import Callable
 from collections.abc import Sequence
 from concurrent.futures import ThreadPoolExecutor
 from pkgutil import walk_packages
+from types import ModuleType
 from typing import Dict
 from typing import List
 from typing import Set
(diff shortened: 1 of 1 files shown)


doctor.py
```diff
@@ -1,55 +1,68 @@
...
-        print(f"repo {repo} not supported yet!")
-        return 1
 
+def load_checks(context: Dict[str, str], match_tags: Set[str]) -> List[Check]:
     checks = []
     for module_finder, name, ispkg in walk_packages((f'{context["reporoot"]}/devenv/checks',)):
         module = module_finder.find_spec(name).loader.load_module(name)  # type: ignore
         assert isinstance(module.name, str)
         assert isinstance(module.tags, set)
+        assert hasattr(module, "check")
```
mattgauntseo-sentry, line 39: Optional: This is just an idea, but moving the assertions into the Check __init__ function would simplify the load_checks logic and I think be a little easier to follow the code

For example:

```
checks.append(Check(module))



class Check:
    def __init__(module):
        assert isinstance(module.name, str)
        name = module.name
        
        assert isinstance(module.tags, set)
        tags = module.tags
        
        ........
```

doctor.py
```diff
@@ -58,6 +71,38 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
             continue
         print(f"\t❌ check: {check.name}{msg}".expandtabs(4))
         retry.append(check)
+    return retry
```
mattgauntseo-sentry, line 74: NIT: Does it make sense to rename retry to something like `failing_checks`?

doctor.py
```diff
@@ -1,55 +1,68 @@
...
+        if check in skip:
+            print(f"\t⏭️ Skipped {check.name}".expandtabs(4))
+            continue
         futures[check] = executor.submit(check.check)
     for check, future in futures.items():
         results[check] = future.result()
+    return results
 
+
+def examine_results(results: Dict[Check, Tuple[bool, str]]) -> List[Check]:
```
mattgauntseo-sentry, line 65: NIT: Should this be renamed to `filter_failing_results`?

doctor.py
IanWoodard: One problem with that would be the duplicate logic needed to check the tags inside of load_checks since we use the tags to decide whether a check should be filtered out.

mattgauntseo-sentry: Yeah sorry I guess I'd go with:

```
c = Check(module)
if match_tags && c.tags.issubset(match_tags):
	checks.append(c)
```

I was largely thinking that Check is responsible to validating input to create a check rather than the loader, but it's questionable if it's any better.

doctor.py
```diff
@@ -43,21 +99,9 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
...
 
-    retry = []
-    for check, result in results.items():
-        ok, msg = result
-        if ok:
-            print(f"\t✅ check: {check.name}".expandtabs(4))
-            continue
-        print(f"\t❌ check: {check.name}{msg}".expandtabs(4))
-        retry.append(check)
+    retry = filter_failing_checks(results)
```
mattgauntseo-sentry, line 104: NIT: Can we change this to failing_checks instead of retry?

retry to me sounds like checks that failed and we want to retry them, which isn't the case here.

IanWoodard: Whoops, yep!

doctor.py
IanWoodard: I see, that works!

joshuarli (approved):

PR #12, opened 2023-10-03
Bad checks shouldn't cause the doctor to break

IanWoodard:
Ideally bad checks wouldn't make it into prod, but if it does happen, we don't want it to break the doctor.


doctor.py
```diff
@@ -42,7 +42,10 @@ def load_checks(context: Dict[str, str], match_tags: Set[str]) -> List[Check]:
     checks = []
     for module_finder, name, ispkg in walk_packages((f'{context["reporoot"]}/devenv/checks',)):
         module = module_finder.find_spec(name).loader.load_module(name)  # type: ignore
-        check = Check(module)
+        try:
+            check = Check(module)
+        except (AttributeError, AssertionError):
```
joshuarli, line 47: can we just capture AssertionError here? any expected attributeerrors would be raised as assertionerrors

IanWoodard: Updated 👍🏻

doctor.py
```diff
@@ -42,7 +44,10 @@ def load_checks(context: Dict[str, str], match_tags: Set[str]) -> List[Check]:
     checks = []
     for module_finder, name, ispkg in walk_packages((f'{context["reporoot"]}/devenv/checks',)):
         module = module_finder.find_spec(name).loader.load_module(name)  # type: ignore
-        check = Check(module)
+        try:
+            check = Check(module)
+        except AssertionError:
+            continue
```
mattgauntseo-sentry, line 50: Optional: Can we log a message that a particular check is being ignored because it's invalid?

Otherwise, it's kind of difficult for checkout authors to know if there is an issue.

mattgauntseo-sentry (approved):

doctor.py
IanWoodard: I totally agree, the reason I didn't add it in here was because of the intentionally broken checks I added. I didn't want those popping up every time. I am planning on using actual tests rather than the current setup so this shouldn't be a problem soon.

PR #14, opened 2023-10-05
Adding more tests and abstracting out more logic

IanWoodard:


doctor.py
```diff
@@ -82,6 +82,23 @@ def filter_failing_checks(results: Dict[Check, Tuple[bool, str]]) -> List[Check]
...
+        "yes",
+        "",
+    }
+
+
+def attempt_fix(check: Check, executor: ThreadPoolExecutor) -> Tuple[bool, str]:
+    future = executor.submit(check.fix)
+    result = future.result()
+    ok, msg = result
+    return ok, msg
```
mattgauntseo-sentry, line 99: Can you do `return result` here? (Not sure if python will let you)

mattgauntseo-sentry (approved):

PR #17, opened 2023-10-06
Handle broken checks

IanWoodard:
Adding logic to handle checks are broken and have exceptions thrown in them.

### DIFF
--- a/doctor.py
+++ b/doctor.py
@@ -75,7 +75,7 @@ def run_checks(
         try:
             results[check] = future.result()
         except Exception as e:
-            results[check] = (False, f"(exception: {e})")
+            results[check] = (False, f"Check threw a runtime exception: {e}")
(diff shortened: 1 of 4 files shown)


doctor.py
```diff
@@ -72,7 +72,10 @@ def run_checks(
             continue
         futures[check] = executor.submit(check.check)
     for check, future in futures.items():
-        results[check] = future.result()
+        try:
+            results[check] = future.result()
+        except Exception as e:
+            results[check] = (False, f"(exception: {e})")
```
mattgauntseo-sentry, line 78: Optional: change the message to explain what happened.

```suggestion
            results[check] = (False, f"Check threw a runtime exception: {e}")
```

mattgauntseo-sentry (approved):

PR #19, opened 2023-10-06
feat: complete devenv sync

joshuarli:
This completes `devenv sync` by having it do everything the post-merge hook does: https://github.com/getsentry/sentry/pull/55486

The goal is to get people to simply say `devenv sync` when the dev environment needs to be refreshed, rather than a specific `make install-py-dev` or `make apply-migrations` etc.

devenv doctor could also run `devenv sync` before doing anything else.

image


bootstrap.py
```diff
@@ -116,16 +117,48 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> int:
...
         proc.run_stream_output(
-            # a new interactive shell is started here so that we get things like updated PATH
-            (shell, "-i", "-e", "-c", "direnv allow ; direnv exec . make bootstrap"),
+            (
+                shell,
+                # interactive shell is used so we can make sure our earlier shellrc
+                # modifications are good
+                "-i",
+                "-e",
+                "-c",
```
asottile-sentry, line 126: we should factor out shells entirely -- this won't work on weird shells (and `-i` / etc. opens us up to people's broken rcfiles)

asottile-sentry (approved):

bootstrap.py
joshuarli: will f/u with this!

PR #21, opened 2023-10-10
Adding better type checking when loading checks

IanWoodard:
As an added layer of protection against incorrect checks, I added additional type checking. Additionally, I added some messages to the assertions to provide better feedback about why a check is invalid.

### DIFF
--- a/doctor.py
+++ b/doctor.py
@@ -4,12 +4,14 @@ import argparse
 from collections.abc import Callable
 from collections.abc import Sequence
 from concurrent.futures import ThreadPoolExecutor
+from inspect import signature
 from pkgutil import walk_packages
 from types import ModuleType
 from typing import Dict
 from typing import List
 from typing import Set
 from typing import Tuple
+import typing
 
 from devenv.lib_check.types import checker
 from devenv.lib_check.types import fixer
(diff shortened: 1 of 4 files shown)


doctor.py
```diff
@@ -41,21 +42,27 @@ def __init__(
...
 
         # Check that the module has the check and fix functions.
-        assert hasattr(module, "check")
-        assert callable(module.check)
+        assert hasattr(module, "check"), "must have a check function"
+        assert callable(module.check), "check should be a function"
+        check_hints = typing.get_type_hints(module.check)
+        assert (
+            check_hints["return"] == Tuple[bool, str]
+        ), "check should return a tuple of (bool, str)"
```
mattgauntseo-sentry, line 59:
```suggestion
        ), "`check(...)` should return a tuple of (bool, str)"
```

doctor.py
```diff
@@ -41,21 +42,27 @@ def __init__(
...
-        assert isinstance(module.tags, set)
+        assert hasattr(module, "tags"), "must have tags"
+        assert isinstance(module.tags, set), "tags should be a set"
         self.tags = module.tags
 
         # Check that the module has the check and fix functions.
-        assert hasattr(module, "check")
-        assert callable(module.check)
+        assert hasattr(module, "check"), "must have a check function"
+        assert callable(module.check), "check should be a function"
```
mattgauntseo-sentry, line 55:
```suggestion
        assert callable(module.check), "the `check `attribute must be a function"
```

doctor.py
```diff
@@ -41,21 +42,27 @@ def __init__(
         module: ModuleType,
     ):
         # Check that the module has the required attributes.
-        assert hasattr(module, "name")
-        assert isinstance(module.name, str)
+        assert hasattr(module, "name"), "must have name"
```
mattgauntseo-sentry, line 45:
```suggestion
        assert hasattr(module, "name"), "missing the `name` attribute"
```

doctor.py
```diff
@@ -41,21 +42,27 @@ def __init__(
...
+        assert isinstance(module.name, str), "name should be a str"
```
mattgauntseo-sentry, line 46:
```suggestion
        assert isinstance(module.name, str), "the `name` attribute should be a str"
```

mattgauntseo-sentry (approved):
Feel free to ignore the suggestions, just an idea - adding any context is a big improvement

mattgauntseo-sentry (approved):
Feel free to ignore the suggestions, just an idea - adding any context is a big improvement

PR #22, opened 2023-10-11
feat: support intel macs

joshuarli:
these fixes were implemented after testing on my intel mac from scratch

also for some reason, macos-13 seems to have changed a bit (and broke in the process), so i've fixed CI as well (mostly by nuking away more homebrew state preinstalled on the runner image)

getsentry doesn't succeed yet - i'll f/u with this soon

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,8 @@ jobs:
           # brew list | xargs brew uninstall --force
           sudo rm -rf \
             /usr/local/bin/brew /usr/local/Homebrew /usr/local/Cellar /usr/local/Caskroom \
-            /usr/local/bin/pydoc* /usr/local/bin/python* /usr/local/bin/2to3* /usr/local/bin/idle3*
+            /usr/local/bin/pydoc* /usr/local/bin/python* /usr/local/bin/2to3* /usr/local/bin/idle3* \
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -17,6 +17,15 @@ jobs:
           python-version: 3.11
       - uses: pre-commit/action@v3.0.0
 
+      - name: remove homebrew
+        run: |
+          # This takes too long (over a minute), so we just rm -rf the important stuff
+          # brew list | xargs brew uninstall --force
```
IanWoodard, line 23: May want to document why we have to remove homebrew

install-devenv.sh
```diff
@@ -42,9 +42,11 @@ if ! [[ -d "${devenv_root}/devenv" ]]; then
 fi
 
 # note: even though this is #!/bin/bash, SHELL isn't modified and so remains the value set by the parent shell
-shellrc="${HOME}/.${SHELL}rc"
+shellrc="${HOME}/.${SHELL##*/}rc"
```
IanWoodard, line 45: nice 👍🏻

IanWoodard (approved):

.github/workflows/test.yml
joshuarli: sure, will do in #163

PR #24, opened 2023-10-17
ref: factor out use of interactive shell so we're not as affected by user's environment

joshuarli:
f/u to https://github.com/getsentry/devenv/pull/19#discussion_r1362664193

> we should factor out shells entirely -- this won't work on weird shells (and -i / etc. opens us up to people's broken rcfiles)


asottile-sentry (approved):
nice!

PR #27, opened 2023-11-03
Setting up craft

IanWoodard:
Setting up craft for automated releases.


.craft.yml
```diff
@@ -0,0 +1,10 @@
+---
+minVersion: 0.34.1
+github:
+  owner: getsentry
+  repo: devenv
+changelogPolicy: auto
```
asottile-sentry, line 6: you can leave these out -- these are the defaults

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: "Prepare Artifacts"
+        run: |
+          pip install wheel
```
asottile-sentry, line 18: you shouldn't need this -- `build` will get it for you

.github/workflows/build.yml
```diff
@@ -0,0 +1,24 @@
...
+          pip install build
+          python -m build --wheel
```
asottile-sentry, line 20: running this with no arguments is preferable since you will get both sdist and wheel

scripts/bump-version.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+sed -i "s/^version =.*/version = $2/" setup.cfg
```
asottile-sentry, line 4: need to double check this will do what you want -- since I don't think this package is using setup.cfg

asottile-sentry (approved):

PR #30, opened 2023-11-08
Making devenv python 3.8 compatible

IanWoodard:
Making devenv compatible with python 3.8 will allow us to use it in Sentry (which still relies on python 3.8).

### DIFF
--- a/devenv/main.py
+++ b/devenv/main.py
@@ -7,7 +7,7 @@ import subprocess
 import time
 from collections.abc import Sequence
 from typing import cast
-from typing import TypeAlias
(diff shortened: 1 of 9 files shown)


devenv/main.py
```diff
@@ -7,7 +7,8 @@
 import time
 from collections.abc import Sequence
 from typing import cast
-from typing import TypeAlias
+
+from typing_extensions import TypeAlias
```
asottile-sentry, line 11: will need to add a dependency on typing_extensions

asottile-sentry (approved):

PR #32, opened 2023-11-13
feat: move venv python to per-repo configuration

joshuarli:
depends on https://github.com/getsentry/sentry/pull/59865 which i'll merge then kick off ci here

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,7 @@ repos:
       - id: mypy
         additional_dependencies:
           - "pytest"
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -20,6 +20,7 @@ repos:
       - id: mypy
         additional_dependencies:
           - "pytest"
+          - "types-toml"
```
asottile-sentry, line 23: can remove this

asottile-sentry (approved):
fixnship

PR #38, opened 2023-11-29
Adding tests for volta

IanWoodard:


tests/volta/test_install.py
```diff
@@ -0,0 +1,76 @@
...
+                call("volta", path=f"{str(tmp_path)}/bin"),
+                call("node", path=f"{str(tmp_path)}/volta/bin"),
+            ]
+
+
+def test_install(tmp_path: str) -> None:
+    with patch.multiple(
+        "devenv.lib.volta",
+        root=str(tmp_path),
+        VOLTA_HOME=f"{str(tmp_path)}/volta",
```
joshuarli, line 33: `str` not necessary in f-strings

tests/volta/test_determine_platform.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
```
joshuarli, line 1: this test file as well as `tests/volta/test_build_url.py` may as well be removed and you should cover it with actually calling build_url in `tests/volta/test_download_and_unpack_archive.py`

i think with unit tests and mocking its easy to make a bunch of tests and i'd like to steer more towards quality over quantity

devenv/lib/volta.py
```diff
@@ -23,49 +23,70 @@ class UnexpectedPlatformError(Exception):
...
+def install_volta(unpack_into: str) -> None:
+    name = determine_platform(unpack_into)
+    if name is None:
+        return
+    download_and_unpack_archive(name, unpack_into)
+
 
-    if (
-        which("volta", path=unpack_into) == f"{unpack_into}/volta"
+def are_volta_and_node_installed(unpack_into: str) -> bool:
```
joshuarli, line 66: i think you've overdone it with breaking code up into too many small functions... this piece of logic is straightforward enough it doesn't warrant the overhead of another function call + a whole test file

joshuarli: the other functions are fine, i'm just proposing you inline this one back

devenv/lib/volta.py
```diff
@@ -23,49 +23,66 @@ class UnexpectedPlatformError(Exception):
...
     elif system == "Darwin":
         suffix = "-aarch64" if platform.machine() == "arm64" else ""
-        name = f"volta-{_version}-macos{suffix}.tar.gz"
+        return f"volta-{_version}-macos{suffix}.tar.gz"
     else:
         raise UnexpectedPlatformError(f"Unexpected OS: {platform.platform()}")
 
-    url = (
+
+def build_url(name: str) -> str:
```
joshuarli, line 45: sorry i missed this one - i don't think we should have a function merely formatting a string and used once

as for the tests, i would assert that `archive.download` was called with the expected url

IanWoodard: Fair point

PR #40, opened 2023-11-29
Adding tests for brew

IanWoodard:
Beginning to add testing for our utility used to install homebrew


devenv/lib/brew.py
```diff
@@ -52,12 +48,29 @@ def install() -> None:
...
+    shellrc = fs.shellrc()
     fs.idempotent_add(
         shellrc,
         f"""
 eval "$({homebrew_bin}/brew shellenv)"
 """,
     )
+
+
+def symlink_brew() -> None:
```
joshuarli, line 62: another too-small function - please just inline this

devenv/lib/brew.py
```diff
@@ -39,6 +33,8 @@ def install() -> None:
             continue
         break
 
+
+def clone_brew() -> None:
```
joshuarli, line 37: same thing here... this really doesn't need to be its own function

tests/brew/test_create_dirs.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
```
joshuarli, line 1: i don't think this entire file is useful - integration tests here provide more than sufficient coverage because if dirs aren't right then 

that said, since we don't run integration mac arm, i'd add an assert that mock_create_dirs was called with the expected /opt/ thing in test_install_brew_not_installed_non_intel_mac

IanWoodard: True, but unit tests can be run locally and are much faster so if a change does break devenv, it is better to know sooner

joshuarli (approved):

PR #42, opened 2023-11-29
Adding tests for GitHub utils

IanWoodard:
Beginning to add testing for our GitHub utility

### DIFF
--- a/devenv/lib/github.py
+++ b/devenv/lib/github.py
@@ -8,10 +8,15 @@ from devenv.constants import home
 from devenv.lib import fs
 from devenv.lib import proc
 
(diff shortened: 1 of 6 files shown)


devenv/lib/github.py
```diff
@@ -40,13 +49,13 @@ def check_ssh_access() -> bool:
     return False
 
 
+def check_ssh_access() -> bool:
+    if CI:
+        return True
+    return check_ssh_authentication() and check_sso_configuration()
```
joshuarli, line 55: may as well skip sso check if external contrib

joshuarli (approved):

PR #44, opened 2023-12-01
test: lib.fs

joshuarli:


tests/lib/test_fs.py
```diff
@@ -0,0 +1,36 @@
...
+
+def test_gitroot() -> None:
+    assert gitroot() == _gitroot
+    assert os.path.isdir(f"{gitroot()}/.git")
+
+
+def test_gitroot_cd() -> None:
+    with chdir(f"{_gitroot}/tests"):
+        assert gitroot() == _gitroot
+        assert os.path.isdir(f"{gitroot()}/.git")
```
asottile-sentry, line 36: I would probably initalize git repos in a temporary directory in the tests -- rather than relying on the git state of the repository itself -- https://github.com/pre-commit/pre-commit/blob/51df34e5fb3e831ebe6a59d9f0004bfa005c0efa/tests/git_test.py#L13-L54

joshuarli: thanks, this is definitely better

tests/lib/test_fs.py
```diff
@@ -0,0 +1,57 @@
...
+    with chdir(tmp_path):
+        subprocess.run(("git", "init"))
+
+        _gitroot = (
+            subprocess.run(
+                ("git", "rev-parse", "--show-toplevel"), capture_output=True
+            )
+            .stdout.decode()
+            .strip()
+        )
```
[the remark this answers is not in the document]
joshuarli: ahhhhhhh, thanks for this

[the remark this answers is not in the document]
joshuarli: also didn't know about samefile, that's nice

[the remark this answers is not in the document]
joshuarli: wow these tests are so nice now haha

tests/lib/test_fs.py
```diff
@@ -0,0 +1,37 @@
...
+    subprocess.run(("git", "init", f"{tmp_path}"))
+
+    with chdir(tmp_path):
+        assert os.path.samefile(tmp_path, gitroot())
+        assert os.path.isdir(f"{gitroot()}/.git")
+
+
+def test_gitroot_cd(tmp_path: pathlib.Path) -> None:
+    subprocess.run(("git", "init", f"{tmp_path}"))
+    subprocess.run(("git", "init", f"{tmp_path}/nested"))
```
asottile-sentry, line 33: is this test supposed to demonstrate that it'll give you the root path even if you're inside of some subdir?  right now it's nested git repos which I don't think is relevant?

joshuarli: uhhhhhhh lemme rework this test, sorry

asottile-sentry (approved):

PR #47, opened 2023-12-04
fix: proc.run combines stdout and stderr by default

joshuarli:
…so that it shows up in sentry if it errors. Someone ran into an issue with `make bootstrap` and the sentry error wasn't useful since stdout+err didn't get put into the exception details: https://sentry.sentry.io/issues/4681931344/events/7ebff3e3a9524b54a949e0bb11b423d2/?project=5723503

I think this is also just a good default and can't remember why I made it an option.

### DIFF
--- a/devenv/bootstrap.py
+++ b/devenv/bootstrap.py
@@ -49,7 +49,7 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> ExitCode:
         # There is a way to perform a headless install but it's more complex
         # (refer to how homebrew does it).
         try:
-            git = proc.run(("xcrun", "-f", "git"), stdout=True)
+            git = proc.run(("xcrun", "-f", "git"))
         except RuntimeError:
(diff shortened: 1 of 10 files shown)


devenv/bootstrap.py
```diff
@@ -49,7 +49,7 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> ExitCode:
         # There is a way to perform a headless install but it's more complex
         # (refer to how homebrew does it).
         try:
-            git = proc.run(("xcrun", "-f", "git"), stdout=True)
+            git = proc.run(("xcrun", "-f", "git"))
```
asottile-sentry, line 52: if this produces a warning on stderr the result will be incorrect

same for the rest of the calls (this is the reason why they are separate and being separate is a _good_ thing) -- will need to find a different way to get it into the error messages

joshuarli: yeah - sorry was working on this late at night / half asleep at the wheel lol

PR #48, opened 2023-12-04
fix: always capture stdout, and additionally surface any stderr in detail

joshuarli:
redo of https://github.com/getsentry/devenv/pull/47

example: https://sentry.sentry.io/issues/4696696613/?project=5723503&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0

i tried some stuff to try and get individual issue groupings but to no avail - can fallback to going to All Events and search by user

### DIFF
--- a/devenv/lib/proc.py
+++ b/devenv/lib/proc.py
@@ -4,8 +4,6 @@ from pathlib import Path
 from subprocess import CalledProcessError
 from subprocess import PIPE
 from subprocess import run as subprocess_run
-from typing import Literal
-from typing import overload
 
 from devenv import constants
 from devenv.constants import home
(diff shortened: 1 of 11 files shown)


devenv/lib/proc.py
```diff
@@ -88,25 +56,27 @@ def run(
     if constants.DEBUG:
         xtrace(cmd)
     try:
-        proc = subprocess_run(cmd, check=True, stdout=_stdout, cwd=cwd, env=env)
-        if _stdout:
+        proc = subprocess_run(
+            cmd, check=True, stdout=PIPE, stderr=PIPE, cwd=cwd, env=env
```
asottile-sentry, line 60: are there no places where we run subprocesses and let the output go directly to the user? (make install-py-dev maybe?)

joshuarli: yeah you're right - i'm gonna rollback the "always capture stdout" part it's not a good idea

joshuarli:
meh

PR #55, opened 2023-12-19
add zstd support to extraction

asottile-sentry:


devenv/lib/archive.py
```diff
@@ -42,7 +47,23 @@ def download(url: str, sha256: str, dest: str = "") -> str:
...
 
 
+class ZstdTarfile(tarfile.TarFile):
+    OPEN_METH = {"zstd": "zstdopen", **tarfile.TarFile.OPEN_METH}
+
+    @classmethod
+    def zstdopen(
+        cls,
+        name: str | None,
+        mode: Literal["r", "a", "w"] = "r",
```
joshuarli, line 57: rawr

joshuarli (approved):

devenv/lib/archive.py
asottile-sentry: xd

asottile-sentry:
actually probably don't need this now :/

PR #58, opened 2023-12-21
feat: autodetect coderoot if already in a git repository

joshuarli:
beta testers are a little annoyed by the coderoot prompt - when you encounter this you need to look up where your sentry's cloned to then type it in exactly

let's just autodetect this for them and make the transition easier

### DIFF
--- a/devenv/main.py
+++ b/devenv/main.py
@@ -70,7 +70,15 @@ def initialize_config(config_path: str, defaults: Config) -> None:
                 # this is a special case used to make the transition from existing
                 # dev environments easier as we can guess the desired coderoot if
                 # devenv is run inside of a git repo
-                breakpoint()
+                try:
+                    reporoot = gitroot()
+                    coderoot = os.path.abspath(f"{reporoot}/..")
+                    print(f"\nWe autodetected a coderoot: {coderoot}")
+                    config.set(section, var, coderoot)
+                    continue
+                except RuntimeError:
+                    pass
+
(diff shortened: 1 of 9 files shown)


devenv/main.py
```diff
@@ -37,6 +37,19 @@ def initialize_config(config_path: str, defaults: Config) -> None:
...
+                # dev environments easier as we can guess the desired coderoot if
+                # devenv is run inside of a git repo
+                try:
+                    reporoot = gitroot()
+                    coderoot = os.path.abspath(f"{reporoot}/..")
+                    print(f"\nWe autodetected a coderoot: {coderoot}")
+                    config.set(section, var, coderoot)
+                    continue
+                except RuntimeError:
+                    pass
```
asottile-sentry, line 51: `try:` / `except:` should wrap the smallest context possible -- and generally be explicit in what exceptions it catches

I think you want:

```python
try:
    reporoot = gitroot()
except RuntimeError:  # TODO: better exception type
    pass
else:
    coderoot = ...
```

joshuarli: oh kinda embarassing but TIL about try except else - thanks!

asottile-sentry: oh! it's so good!

joshuarli: i don't know how to go about a more specific exception type here because stdout doesn't make it through to  `detail`, and it's not a specific git returncode :(

asottile-sentry (approved):

PR #59, opened 2023-12-21
feat: manage colima ourselves

joshuarli:
some people have been running into issues with bad colima versions, with 0.6.2 being the known good one 

i haven't found a way to add/reference a local formula file to Brewfile, and i'm thinking the best path forward is to just let devenv install colima which is just a go binary which just works per my testing

really don't like how brew doesn't let you easily install a particular version of something and i don't want people to get broken by updating colima

### DIFF
--- a/devenv/constants.py
+++ b/devenv/constants.py
@@ -26,6 +26,7 @@ user_environ: typing.Mapping[str, str] = os.environ.copy()
 cache_root = f"{home}/.cache/sentry-devenv"
 config_root = f"{home}/.config/sentry-devenv"
 root = f"{home}/.local/share/sentry-devenv"
+bin_root = f"{root}/bin"
 src_root = f"{root}/devenv"
 pythons_root = f"{root}/pythons"
 
--- a/devenv/constants.py
+++ b/devenv/constants.py
@@ -30,14 +30,15 @@ bin_root = f"{root}/bin"
 src_root = f"{root}/devenv"
 pythons_root = f"{root}/pythons"
 
-
-if INTEL_MAC:
-    homebrew_repo = "/usr/local/Homebrew"
-    homebrew_bin = "/usr/local/bin"
-else:  # FIXME: elif ARM_MAC
-    homebrew_repo = "/opt/homebrew"
+if DARWIN:
+    if MACHINE == "x86_64":
+        homebrew_repo = "/usr/local/Homebrew"
+        homebrew_bin = "/usr/local/bin"
+    else:
+        homebrew_repo = "/opt/homebrew"
+        homebrew_bin = f"{homebrew_repo}/bin"
+else:
+    homebrew_repo = f"{home}/linuxbrew/.linuxbrew"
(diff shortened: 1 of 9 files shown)


devenv/lib/colima.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def download(name: str, into: str) -> None:
+    url = (
+        "https://github.com/abiosoft/colima/releases/download/"
+        f"v{_version}/{name}"
+    )
+
+    archive.download(url, _sha256[name], dest=f"{into}/colima")
+    os.chmod(f"{into}/colima", 0o775)
```
asottile-sentry, line 43: one thing I noticed (and we need to fix this in general -- not just here) -- is a partially successful download will break the share data for ever

for instance:
- a partial download of the `colima` file
- a partially extracted archive
- a misconfigured python which pulls the wrong archive (will place the `3.x.x` directory that can't be cleaned up later)

we should probably:
1. download to a temporary location adjacent to the target
2. atomically move to the target

this is tricky to do for directories in a cross platform way unfortuantely -- for files it's easy (`os.replace` is idempotent / atomic)

joshuarli: did it for files

devenv/lib/archive.py
```diff
@@ -21,23 +22,24 @@ def download(url: str, sha256: str, dest: str = "") -> str:
...
-
-    if not secrets.compare_digest(checksum.hexdigest(), sha256):
-        os.remove(dest)
-        raise RuntimeError(
-            f"checksum mismatch for {url}:\n"
-            f"- got: {checksum.hexdigest()}\n"
-            f"- expected: {sha256}\n"
-        )
+
+        with tempfile.NamedTemporaryFile(delete=False) as tmpf:
```
asottile-sentry, line 26: specifically this tempfile has to be adjacent to where it will end up -- otherwise it could attempt to rename across partitions -- you can use `dir=...` to control that

joshuarli: oh, i suppose rename atomicity is only guaranteed within the same fs, that makes sense

asottile-sentry: yeah -- it's even an error when attempting a rename across filesystems

asottile-sentry (approved):

devenv/constants.py
```diff
@@ -26,17 +26,19 @@
...
+if DARWIN:
+    if MACHINE == "x86_64":
+        homebrew_repo = "/usr/local/Homebrew"
+        homebrew_bin = "/usr/local/bin"
+    else:
+        homebrew_repo = "/opt/homebrew"
+        homebrew_bin = f"{homebrew_repo}/bin"
+else:
+    homebrew_repo = f"{home}/linuxbrew/.linuxbrew"
     homebrew_bin = f"{homebrew_repo}/bin"
```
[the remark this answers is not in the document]
joshuarli: yeah i agree, i'm gonna walk this stuff back and just have docker runtime in linux ci

asottile-sentry (approved):

PR #61, opened 2024-01-08
test: bootstrap

joshuarli:


IanWoodard (changes requested):
Could we stick to the same directory pattern used for the other tests or update those to match?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,136 @@
...
+    ) as mock_proc_run:
+        coderoot = f"{tmp_path}/coderoot"
+        bootstrap.main(coderoot=coderoot, argv=("sentry",))
+        mock_add_to_known_hosts.assert_called_once()
+        mock_generate_and_configure_ssh_keypair.assert_called_once()
+        mock_brew_install.assert_called_once()
+        mock_volta_install.assert_called_once()
+        mock_direnv_install.assert_called_once()
+        mock_colima_install.assert_called_once()
+        mock_limactl_install.assert_called_once()
```
IanWoodard, line 47: Nit: Could we assert that these were called with specific inputs?

IanWoodard (approved):
One nit, but otherwise lgtm

tests/test_bootstrap.py
joshuarli: they don't take any arguments

PR #62, opened 2024-01-09
test: sync

joshuarli:


tests/test_sync.py
```diff
@@ -0,0 +1,194 @@
...
+        "devenv.sync.run_procs", side_effect=[True, True, True]
+    ) as mock_run_procs:
+        reporoot = f"{tmp_path}/sentry"
+        venv = f"{tmp_path}/sentry/.venv"
+
+        os.makedirs(f"{reporoot}/devenv")
+        os.makedirs(venv)
+
+        with open(f"{venv}/pyvenv.cfg", "w") as f:
+            f.write("version = 3.8.16")
```
IanWoodard, line 46: Should we be hard coding this? It makes it harder to maintain

tests/test_sync.py
```diff
@@ -0,0 +1,194 @@
...
```
IanWoodard, line 132: Should we be hard coding this? It makes it harder to maintain

IanWoodard (approved):
One suggestion (in two places), other than that, I think we could abstract some of the code to reduce the amount of duplicate code but otherwise lgtm.

tests/test_sync.py
joshuarli: this is fine and never needs to change because the main purpose here is to demonstrate venv recreation

PR #64, opened 2024-01-10
test: direnv

joshuarli:


tests/lib/test_direnv.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+
+import os
+from unittest import mock
+from unittest.mock import call
+from unittest.mock import patch
+
+
+# imo this is enough coverage for such a trivial thing
```
IanWoodard, line 9: While I mostly agree with the comment about coverage, could we add a clear description or docstring to `test_install` that outlines what this test specifically aims to verify? This would make it easier for anyone reading in the future to understand the purpose and scope of this test.

IanWoodard (approved):
Lgtm besides one comment

tests/lib/test_direnv.py
joshuarli: sure

PR #67, opened 2024-01-12
ref: replace assert_has_calls with comparing against ordered mock_calls

joshuarli:
this finishes replacing assert_has_calls (which can be true on subsets of calls) to strictly comparing against all calls done in-order


IanWoodard (approved):
👍🏻

PR #68, opened 2024-01-16
doc: update readme with some technical info

joshuarli:


README.md
```diff
@@ -13,6 +13,35 @@ bash install-devenv.sh
 **If you are setting up a new development environment, run `devenv bootstrap` after installing.**
 
 
+
+## what is it / technical overview
```
IanWoodard, line 17: Since this isn't a question, should we change it to: `what it is`?

README.md
```diff
@@ -13,6 +13,35 @@ bash install-devenv.sh
 **If you are setting up a new development environment, run `devenv bootstrap` after installing.**
```
IanWoodard, line 13: I know this is unrelated to this PR, but I think we should explain that devenv bootstrap should be run in specific repos.

IanWoodard (approved):
A few nits, but otherwise lgtm

README.md
joshuarli: i'll change it to just `technical overview`

README.md
joshuarli: i'll change this to "If you are setting up a new laptop, run `devenv bootstrap` after installing."

it's a one-time thing for new engineers, not something that should be run in specific repos

IanWoodard (approved):

PR #71, opened 2024-01-30
feat: lib.venv and multiple configurable venvs (ops POC part 1/3)

joshuarli:
this is a new `lib.venv` which'll make it easy to script venv management for doctor and for people implementing repo-local sync

the idea going forward is that we should let people mostly script these things rather than contribute to devenv which we then have to release new versions, which is definitely not scalable

see https://github.com/getsentry/ops/pull/9296 which has a new devenv/config.ini and devenv/sync.py which describe two separate virtualenvs with different python versions (salt doesn't work on 3.11, but sentry-kube can be on it)

---

i plan on adding some new unit and integration tests after all 3 parts/PRs are merged


devenv/lib/venv.py
```diff
@@ -0,0 +1,165 @@
...
+from devenv.constants import bin_root
+from devenv.constants import venvs_root
+from devenv.lib import config
+from devenv.lib import fs
+from devenv.lib import proc
+
+VENV_OK = 1
+VENV_VERSION_MISMATCH = 2
+VENV_NOT_PRESENT = 3
+VENV_NOT_CONFIGURED = 4
```
IanWoodard, line 17: This is a prime case for using enums, it would enhance readability and reduce potential errors.

devenv/lib/venv.py
```diff
@@ -0,0 +1,165 @@
...
+    return (
+        venv_dir,
+        venv["python"],
+        f"{reporoot}/{venv['requirements']}",
+        editable_paths,
+        bins,
+    )
+
+
+def sync(
```
IanWoodard, line 76: It would be nice to have docstrings for functions to get a quick summary of what they are doing

devenv/lib/venv.py
joshuarli: What part of this isn't readable? I fail to see how using enums would be beneficial. If anything it's just more bloated.

devenv/lib/venv.py
joshuarli: sure thing, i can add em in https://github.com/getsentry/devenv/pull/80

devenv/lib/venv.py
IanWoodard: I understand where you are coming from, I prefer using enums for the semantics. Enums provide a clear indication that the values are part of a single group of related values, status codes in this case. This makes the code more self-documenting and readable in my opinion.

joshuarli: alright, fair enough

IanWoodard (approved):

PR #74, opened 2024-02-06
feat: multiple configurable venvs

joshuarli:
this expands on lib.venv by adding new stuff that people can use for scripting venv management and syncing

the idea going forward is that we should let people mostly script these things rather than contribute to `devenv` which we then have to release new versions, which is definitely not scalable

see https://github.com/getsentry/ops/pull/9296 which has a new `devenv/config.ini` and `devenv/sync.py` which describe two separate virtualenvs with different python versions (salt doesn't work on 3.11, but sentry-kube can be on it)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.egg-info/
 __pycache__/
 venv/
+build/
 
 # dirty hacks
 trash/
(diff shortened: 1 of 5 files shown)


joshuarli:
superseded by https://github.com/getsentry/devenv/pull/71 which i've rolled this PR up into

PR #75, opened 2024-02-07
feat: call repo-local sync if available (ops POC part 3/3)

joshuarli:
this lets people write their own sync (example: https://github.com/getsentry/ops/pull/9296/files#diff-9ee0980647c538c23af5cafa555f7d3894165fbacef650c5e191a90e5e24048aR1) which is the vision i described in https://github.com/getsentry/devenv/pull/71

also lint+test are now 3.11

---

i plan on adding some new unit and integration tests after all 3 parts/PRs are merged


devenv/sync.py
```diff
@@ -79,12 +80,20 @@ def run_procs(
...
+            "sync", f"{reporoot}/devenv/sync.py"
+        )
+        module = importlib.util.module_from_spec(spec)  # type: ignore
+        spec.loader.exec_module(module)  # type: ignore
+        return module.main(context)  # type: ignore
+
+    # past this point is legacy code that should only be run for sentry/getsentry
+    # eventually to be moved to sentry/devenv/sync.py
+    assert repo in {"sentry", "getsentry"}
+
```
IanWoodard, line 96: Add a TODO or FIXME so we can more easily remember to come back to this?

IanWoodard (approved):

devenv/sync.py
joshuarli: nah, i plan on doing this once it's shipped

PR #76, opened 2024-02-07
feat: move gcloud sources to repo configuration

joshuarli:
this is so that people can self serve version upgrades in the future without needing to change devenv code


joshuarli:
superseded https://github.com/getsentry/devenv/pull/78

PR #80, opened 2024-02-13
test: add tests for ops poc before i expand on it further

joshuarli:
f/u to https://github.com/getsentry/devenv/pull/75


joshuarli:
outdated

PR #81, opened 2024-02-14
feat: repo-local colima

joshuarli:
so... people are starting to run into some problems with colima 0.6.2 and we should update to latest anyways (some CVEs are fixed)

i've moved this to the new repo-local approach (see https://github.com/getsentry/devenv/pull/78) which is better and more flexible

running all newest devservices on macos 14.3 works on my M1

after this, planning on releasing a new version along with adding to sentry devenv/config.ini


IanWoodard (approved):
Once we get the tests passing, looks good to go

joshuarli:
after some discussion with @asottile-sentry i'm removing the sync and bootstrap unittests as they don't really do much beyond calling functions, and unittests are useful for individual functions

i'll follow up soon with a unittest for run_procs but want to get this out today

devenv/lib/config.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
```
joshuarli, line 1: unittest coverage for this will be addressed in https://github.com/getsentry/devenv/pull/80

asottile-sentry (approved):

PR #84, opened 2024-02-16
test: remove macos-13 as it no longer works with colima 0.6.8

joshuarli:
see https://github.com/getsentry/devenv/pull/83

i need this to unblock a new release that'll get sentry people onto newer better colima 0.6.8

however... should we alternatively run macos-13 + colima 0.6.2 so we can have integration coverage albeit on outdated platform + software?

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -18,6 +18,8 @@ env:
   DEBUG: "1"
 
 jobs:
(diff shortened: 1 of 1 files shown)


IanWoodard (commented):
We definitely shouldn't remove our only integration test, I would opt for keeping it around (with the older version of colima) till we can get a macos-14 runner.

joshuarli:
superseded by https://github.com/getsentry/devenv/pull/85

PR #88, opened 2024-02-20
ref: remove legacy venv

joshuarli:
this removes the legacy sync stuff used for sentry/getsentry! also adds missing unittest coverage

will follow up with adding integration tests for devenv.lib.sync since sentry isn't using it (see https://github.com/getsentry/sentry/pull/65502)


devenv/bootstrap.py
```diff
@@ -98,7 +98,13 @@ def main(coderoot: str, argv: Sequence[str] | None = None) -> ExitCode:
             # git@ clones forces the use of cloning through SSH which is what we want,
             # though CI must clone open source repos via https (no git authentication)
             additional_flags = (
-                ("--depth", "1", "https://github.com/getsentry/sentry")
+                (
+                    "--depth",
+                    "1",
+                    "--single-branch",
+                    f"--branch={os.environ['SENTRY_BRANCH']}",
```
joshuarli, line 105: making it easier to test on sentry branches

dbennett-sentry (approved):

PR #94, opened 2024-02-26
fix: stall

joshuarli:
this is so weird

this stalls in sentry's sync script but if i early return (and then let bootstrap complete, which surprisingly it does successfully as make bootstrap will also try to install sentry) and ssh in and run it manually it succeeds

`run_procs("sentry", "/home/runner/code/sentry", "/home/runner/code/sentry/.venv", ( ("python dependencies", ("make", "install-py-dev")),))`

so... i need to get into the environment before bootstrap

ok. was able to reproduce (although, had to use run_procs from sentry-devenv.sync 1.2.3). the first time, it stalls and doesn't do anything other than print this twice:

```
[notice] A new release of pip is available: 23.3.1 -> 24.0
[notice] To update, run: pip install --upgrade pip
```

interestingly if I just repeat it, it prints that once, then runs to completion.
also makefile's output is missing since run_procs hides stdout to be neat


joshuarli:
https://github.com/getsentry/sentry/pull/66655

PR #100, opened 2024-03-14
ref: factor out root constants except bin_root

joshuarli:
will follow up with factoring out bin_root as it's not trivial to do

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,50 +1,21 @@
 from __future__ import annotations
 
-import os
-import subprocess
-from unittest.mock import call
 from unittest.mock import patch
 
 from devenv import main
-from tests.utils import chdir
 
 
-def test(tmp_path: str) -> None:
-    coderoot = f"{tmp_path}/coderoot"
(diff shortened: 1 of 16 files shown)


tests/test_main.py
```diff
@@ -1,50 +1,21 @@
...
-        ) as os_makedirs:
-            main.devenv(("/path/to/argv0", "sync"))
-
-            # this time, we should have early returned from initialize_config
-            # so only the os.makedirs call to coderoot should have happened
-            # otherwise we would also have a call(configroot, exist_ok=True)
-            assert os_makedirs.mock_calls == [call(coderoot, exist_ok=True)]
-            assert mock_sync.mock_calls == [
-                call({"repo": repo, "reporoot": reporoot}, [])
-            ]
```
asottile-sentry, line 50: isn't this part still important / useful to test?

joshuarli: oh i deleted this with the intention of reworking it - forgot to do so

asottile-sentry (approved):

PR #77, opened 2024-02-07
feat: ops bootstrap

joshuarli:
i plan on putting this up for review once i make further progress on top of the ops poc

PR #101, opened 2024-03-18
ref: reporoot/devenv to reporoot/.devenv

joshuarli:
this is so `python -m devenv` can work inside reporoot

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-22.04
     timeout-minutes: 60
     env:
-      SENTRY_BRANCH: master
+      SENTRY_BRANCH: upgrade-devenv
       SNTY_DEVENV_BRANCH:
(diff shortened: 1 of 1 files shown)


devenv/doctor.py
```diff
@@ -173,12 +170,8 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
 
     match_tags: set[str] = set(args.tag if args.tag else ())
 
-    repo = context["repo"]
-    if repo not in {"sentry", "getsentry", "devenv"}:
```
joshuarli, line 177: repo not supported yet is outdated, i've been getting rid of it

last one is removed in https://github.com/getsentry/devenv/pull/77

asottile-sentry:
won't this create backward/forward incompatibility issues?

this needs to be a 3 step (at least) process

dbennett-sentry (commented):
Backward compatible `.devenv` support has been added to main. I think we can move forward with updating the tests.

PR #104, opened 2024-03-25
ref: pin global devenv to 1.4.0

joshuarli:
we shouldn't be installing the global devenv from `main` which is not very stable in between releases... pin it to 1.4.0 for now so new hires and anyone else running install-devenv.sh don't get stuck on some potentially bad commit

### DIFF
--- a/install-devenv.sh
+++ b/install-devenv.sh
@@ -88,7 +88,9 @@ parseopt() {  # argument (and environment-var) processing
   appname="sentry-devenv"
   # control install behavior
   SNTY_DEVENV_REPO="${SNTY_DEVENV_REPO:-https://github.com/getsentry/devenv.git}"
-  SNTY_DEVENV_BRANCH="${1:-${SNTY_DEVENV_BRANCH:-main}}"
+  # TODO: we need the release action to tag the release commit
+  # 1.4.0
+  SNTY_DEVENV_BRANCH="${1:-${SNTY_DEVENV_BRANCH:-bdd30f96212769ddaec717192750730bb3350402}}"
   SNTY_DEVENV_HOME="${SNTY_DEVENV_HOME:-$XDG_DATA_HOME/$appname}"
   SNTY_DEVENV_CACHE="${SNTY_DEVENV_CACHE:-$XDG_CACHE_HOME/$appname}"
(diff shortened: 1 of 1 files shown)


IanWoodard (commented):
Should we just use another "release" branch to denote stable changes?

joshuarli:
> Should we just use another "release" branch to denote stable changes?

what's wrong with pinning global devenv version?

IanWoodard:
> > Should we just use another "release" branch to denote stable changes?
> 
> what's wrong with pinning global devenv version?

It is harder to track which changes are included in the release (vs diffing the branches) as well as a less clear history. However, both approaches are valid, I just thought I would bring it up to hear your thoughts.

install-devenv.sh
```diff
@@ -88,7 +88,9 @@ parseopt() {  # argument (and environment-var) processing
   appname="sentry-devenv"
   # control install behavior
   SNTY_DEVENV_REPO="${SNTY_DEVENV_REPO:-https://github.com/getsentry/devenv.git}"
-  SNTY_DEVENV_BRANCH="${1:-${SNTY_DEVENV_BRANCH:-main}}"
```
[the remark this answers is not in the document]
joshuarli: ah good idea

install-devenv.sh
```diff
@@ -180,7 +180,13 @@ main() {
 
   show "$devenv_python" -m venv --clear "$devenv_venv"
 
-  show "$devenv_venv"/bin/pip install "git+$SNTY_DEVENV_REPO@$SNTY_DEVENV_BRANCH"
+  if [[ "${CI:-}" ]]; then
+    show "$devenv_venv"/bin/pip install "git+$SNTY_DEVENV_REPO@$SNTY_DEVENV_BRANCH"
+  else
+    show "$devenv_venv"/bin/pip install \
+      --index-url https://pypi.devinfra.sentry.io/simple \
+      "sentry-devenv==${SNTY_DEVENV_VERSION}"
```
asottile-sentry, line 188: could even just leave this unpinned so that our *released* version is what gets pulled in

joshuarli: :hmmyes:

asottile-sentry (approved):

PR #111, opened 2024-04-08
fix: regression in pythons root

joshuarli:
just noticed an unfortunate regression in 3b27f121788bcdbb1a4528061a588fc78da7bd63 where i misinlined pythons_root to {home}/pythons and not {root}/pythons, lol

could have easily been avoided with a simple unittest which i've added

### DIFF
--- a/tests/test_pythons.py
+++ b/tests/test_pythons.py
@@ -1,0 +1,14 @@
+from __future__ import annotations
+
+from unittest.mock import patch
+
+from devenv import pythons
+from devenv.constants import root
+
+
+def test_get() -> None:
+    with patch("devenv.lib.archive.download"), patch(
+        "devenv.lib.archive.unpack"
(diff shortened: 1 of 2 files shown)


tests/test_pythons.py
```diff
@@ -0,0 +1,14 @@
...
+from devenv import pythons
+from devenv.constants import root
+
+
+def test_get() -> None:
+    with patch("devenv.lib.archive.download"), patch(
+        "devenv.lib.archive.unpack"
+    ), patch("os.path.exists", side_effect=[False, True]):
+        path = pythons.get("3.12.0", "foo", "bar")
+        assert path == f"{root}/pythons/3.12.0/python3"
```
IanWoodard, line 14: This doesn't match the function's implementation

IanWoodard (approved):

PR #113, opened 2024-04-12
feat: Split bootstrap and fetch commands; add support for fetching any repo

dbennett-sentry:
This pr separates bootstrapping the devenv tool from fetching workspace repositories. This has also added the ability to fetch any arbitrary github repo.

### DIFF
--- a/devenv/fetch.py
+++ b/devenv/fetch.py
@@ -17,7 +17,7 @@ ExitCode: TypeAlias = "str | int | None"
 
 
 def _fetch(
-    coderoot: str, repo: str, noauth: bool = False, sync: bool = True
+    coderoot: str, repo: str, auth: bool = True, sync: bool = True
 ) -> None:
     org, slug = repo.split("/")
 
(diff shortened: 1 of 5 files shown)


devenv/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+from devenv.constants import EXTERNAL_CONTRIBUTOR
+from devenv.constants import homebrew_bin
+from devenv.lib import proc
+
+help = "Fetches a respository"
+ExitCode: TypeAlias = "str | int | None"
+
+
+def _fetch(
+    coderoot: str, repo: str, noauth: bool = False, sync: bool = True
```
joshuarli, line 20: nit but i'd change `noauth=False` to `auth=False` so it's more obvious to callers that a repository is private

devenv/fetch.py
```diff
@@ -0,0 +1,120 @@
...
+    additional_args = (
+        (
+            "--depth",
+            "1",
+            "--single-branch",
+            f"--branch={os.environ['SENTRY_BRANCH']}",
+            f"https://github.com/{repo}",
+        )
+        if noauth
+        else (f"git@github.com:{repo}",)
```
joshuarli, line 41: please bring back my comment about this

devenv/fetch.py
dbennett-sentry: Yeah, yeah.. I should avoid negatives in variable names.. Good point.

devenv/fetch.py
dbennett-sentry: Umm.. Yep. Unintentional.

joshuarli (approved):
overall this looks great! some followups:

new release needs to be cut (manually run the release workflow here https://github.com/getsentry/devenv/actions/workflows/release.yml)

docs should to be updated (yes, despite bootstrap saying you need to followup with a fetch): https://develop.sentry.dev/environment/#setup

PR #116, opened 2024-04-19
Added new repository and context files; a couple of changes to tests

dbennett-sentry:
These changes will be used in other PRs.

### DIFF
--- a/ci/devenv-bootstrap.sh
+++ b/ci/devenv-bootstrap.sh
@@ -14,6 +14,7 @@ EOF
 
 : PATH: "$PATH"
 
+cd "$HOME"
 # note: colima will be used and is necessary for a docker runtime on
 #       macos GitHub runners
 devenv bootstrap
--- a/devenv/lib/context.py
+++ b/devenv/lib/context.py
@@ -1,0 +1,14 @@
+from __future__ import annotations
+
+from typing import TypeAlias
+from typing import TypedDict
+
+from devenv.lib.repository import Repository
+
+Path: TypeAlias = str
+
+
+class Context(TypedDict):
+    config_path: Path
+    code_root: Path
+    repo: Repository | None
--- a/devenv/lib/repository.py
+++ b/devenv/lib/repository.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+import os.path
+from configparser import ConfigParser
+
+from devenv.lib.config import get_config
+from devenv.lib.fs import ensure_binroot
+
+
+class Repository:
+    path: str
+    name: str
+    config_path: str
+
+    def __init__(self, root: str) -> None:
+        self.path = root
+        self.name = root.split("/")[-1]
+
+        # .devenv may exist but not be config_path because .devenv/bin is hard-coded, so we check for config.ini
+        if os.path.exists(f"{root}/.devenv/config.ini"):
+            self.config_path = f"{root}/.devenv"
+        # devenv _may_ exist but not be config_path because devenv has a devenv module, so we check for config.ini
+        elif os.path.exists(f"{root}/devenv/config.ini"):
+            self.config_path = f"{root}/devenv"
+        else:
+            # new default config_path is .devenv
+            self.config_path = f"{root}/.devenv"
+
+    def __repr__(self) -> str:
+        return self.config_path
+
+    def __str__(self) -> str:
+        return f"Repository: {self.name}"
+
+    def ensure_binroot(self) -> None:
+        ensure_binroot(self.path)
+
+    def config(self) -> ConfigParser:
+        return get_config(f"{self.config_path}/config.ini")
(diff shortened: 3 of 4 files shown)


devenv/lib/context.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+from typing import TypeAlias
+from typing import TypedDict
+
+from devenv.lib.repository import Repository
+
+Path: TypeAlias = str
```
asottile-sentry, line 8: I avoid type aliases -- especially ones line this since they don't actually increase type safety here (but create indirection)

this is especially confusing because there is a common `Path` type from pathlib

but also type aliases are transitive so they don't help with enforcing stricter typing:

```python
Path: TypeAlias = str
y = 'foo'
reveal_type(y)  # str
x: Path = y  # not an error
```

there's `NewType` -- but I haven't found a usecase beyond opaque credential typing that isn't a huge pain without much gain

devenv/lib/repository.py
```diff
@@ -0,0 +1,39 @@
...
+from configparser import ConfigParser
+
+from devenv.lib.config import get_config
+from devenv.lib.fs import ensure_binroot
+
+
+class Repository:
+    path: str
+    name: str
+    config_path: str
```
asottile-sentry, line 13: these annotations here don't do anything (and in general should be left off) -- mypy can infer the types from `__init__` and it's more likely these get out of sync (or break in mysterious ways since the declarations aren't checked for actual presence after `__init__`) https://www.notion.so/sentry/sentry-typing-guidelines-best-practices-471d5aacf262476188982ce70f909e00?pvs=4#4d926cb8ed3647f79f0e9164671705c1

devenv/lib/repository.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, root: str) -> None:
+        self.path = root
+        self.name = root.split("/")[-1]
```
asottile-sentry, line 17: `os.path.basename(root)`

devenv/lib/repository.py
```diff
@@ -0,0 +1,39 @@
...
+
+        # .devenv may exist but not be config_path because .devenv/bin is hard-coded, so we check for config.ini
+        if os.path.exists(f"{root}/.devenv/config.ini"):
+            self.config_path = f"{root}/.devenv"
+        # devenv _may_ exist but not be config_path because devenv has a devenv module, so we check for config.ini
```
asottile-sentry, line 22: I think this `if` block can be left off entirely -- it'll be handled by `else` in every case I can think of -- except maybe a symlink where *both* exist 🤔

devenv/lib/repository.py
```diff
@@ -0,0 +1,39 @@
...
+        elif os.path.exists(f"{root}/devenv/config.ini"):
+            self.config_path = f"{root}/devenv"
+        else:
+            # new default config_path is .devenv
+            self.config_path = f"{root}/.devenv"
+
+    def __repr__(self) -> str:
+        return self.config_path
```
asottile-sentry, line 30: `__repr__` should contain a programmer-specific representation of an object, for example `Repository('whatever/path')` -- if this class extends `typing.NamedTuple` you'll get all this for free

devenv/lib/repository.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __str__(self) -> str:
+        return f"Repository: {self.name}"
```
asottile-sentry, line 33: generally you shouldn't override `__str__` unless you want something to be directly convertable to a string -- this looks like a repr but not in the proper format -- I'd just remove this personally

devenv/lib/context.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Context(TypedDict):
```
asottile-sentry, line 11: might be nice to change this to an actual class rather than a dict -- up to you!

ci/devenv-bootstrap.sh
```diff
@@ -14,6 +14,7 @@ EOF
 
 : PATH: "$PATH"
 
+cd "$HOME"
```
asottile-sentry, line 17: curious why this bit is needed 🤔

devenv/lib/context.py
dbennett-sentry: Yeah, this didn't provide any level of edit-time insight that I was hoping for. Will remove after I get all feedback.

devenv/lib/repository.py
dbennett-sentry: This is here for migration purposes. When we move the sync.py from devenv to .devenv a config.ini may have already been created. As the config wouldn't be checked into git, it'd remain until manually deleted by the user.

ci/devenv-bootstrap.sh
dbennett-sentry: This is needed because the test executes inside a test repository. This confounds the coderoot detection.  I can now override it with the -d argument I had added before but that makes the run during the integration test less real-world.

IanWoodard (approved):

PR #117, opened 2024-04-19
Move config initialization to bootstrap

dbennett-sentry:


tests/test_main.py
```diff
@@ -6,26 +6,28 @@
...
 
 from devenv import main
+from devenv.lib import config
+from devenv.lib.config import DEFAULT_CONFIG
 from tests.utils import chdir
 
 
-def test(tmp_path: str) -> None:
-    with patch("devenv.main.CI", True), patch(
+def _test(tmp_path: str) -> None:
```
asottile-sentry, line 14: this disables the test -- not sure if that was intentional

tests/test_main.py
```diff
@@ -6,26 +6,28 @@
...
+    with patch("devenv.constants.CI", True), patch(
         "devenv.bootstrap.main"
     ) as mock_bootstrap, patch("devenv.sync.main") as mock_sync:
         config_path = f"{tmp_path}/.config/sentry-devenv/config.ini"
         coderoot = f"{tmp_path}/code"
-        main.DEFAULT_CONFIG["devenv"]["coderoot"] = coderoot
+        DEFAULT_CONFIG["devenv"]["coderoot"] = coderoot
```
asottile-sentry, line 20: this irreversibly changes the global variable -- you probably want:

```python
with mock.patch.dict(DEFAULT_CONFIG["devenv"], {"coderoot": coderoot}):
   ...
```

this will clean itself up after the block is exited

devenv/bootstrap.py
```diff
@@ -12,15 +12,24 @@
...
 
-def main(coderoot: str, argv: Sequence[str] | None = None) -> ExitCode:
+def main(
+    config_path: str, coderoot: str, argv: Sequence[str] | None = None
+) -> ExitCode:
     parser = argparse.ArgumentParser(description=help)
     parser.parse_args(argv)
 
+    default_config: Config = {**DEFAULT_CONFIG}
+    default_config["devenv"].update({"coderoot": coderoot})
```
asottile-sentry, line 30: ah while looking at the test this reminds me of another problem here -- `{**d}` is a shallow copy:

```pycon
>>> DEFAULT_CONFIG = {"whatever": 1, "devenv": {"coderoot": "wat"}}
>>> other = {**DEFAULT_CONFIG}
>>> other["devenv"]["coderoot"] = "something-else"
>>> DEFAULT_CONFIG  # oops!
{'whatever': 1, 'devenv': {'coderoot': 'something-else'}}
```

probably needs a deepcopy -- or `DEFAULT_CONFIG` should maybe be a function which *returns* a fresh dict each time?

asottile-sentry (commented):
the rest is mostly rearrangement so I didn't read that part (let me know if I should put more scrutiny there!) -- but the global mutation and test disable should definitely be fixed before merge

tests/test_main.py
dbennett-sentry: It's re-enabled in a later PR after I add some additional fixes.

tests/test_main.py
dbennett-sentry: Ah.. True.  I've removed all of these shenanigans in a later PR.

devenv/bootstrap.py
dbennett-sentry: This line is gone, however, thanks for the reminder to watch my nested structures. I think DEFAULT_CONFIG can be deleted now -- I'll do that as the action here.

tests/test_main.py
dbennett-sentry: New test doesn't contain any mutation on global objects.

`def test(tmp_path: str) -> None:
    with patch("devenv.constants.CI", True):
        config_path = f"{tmp_path}/.config/sentry-devenv/config.ini"
        coderoot = f"{tmp_path}/code"

        assert not os.path.exists(config_path)
        main.devenv(
            ("/path/to/argv0", "bootstrap", "-d", f"coderoot:{coderoot}"),
            config_path,
        )
        assert os.path.exists(config_path)

        config = read_config(config_path)
        assert config.get("devenv", "coderoot") == coderoot`

devenv/bootstrap.py
dbennett-sentry: Constant removed fully.

devenv/lib/config.py
```diff
@@ -2,11 +2,97 @@
...
+        "coderoot",
+        "# please enter the root directory you want to work in",
+        _path_formatter,
+        lambda: "~/code",
+    )
+}
+
+DEFAULT_CONFIG: Config = {"devenv": {"coderoot": "~/code"}}
+
+Config: TypeAlias = "dict[str, dict[str, str | None]]"
```
IanWoodard, line 38: Is the string literal necessary here?

IanWoodard (approved):

tests/test_main.py
```diff
@@ -6,26 +6,28 @@
...
 
-        main.devenv(("/path/to/argv0", "bootstrap"), config_path)
+        config.initialize_config(config_path, DEFAULT_CONFIG)
 
-        assert mock_bootstrap.mock_calls == [call(coderoot, [])]
+        main.devenv(("/path/to/argv0", "bootstrap"), config_path)
```
asottile-sentry, line 24: hmmm needing to pass `argv0` is weird here -- argparse's default mode ignores the first argument already so we can probably fix this up (looks like it'd just be dropping all the `argv0`s in tests and removing a `[1:]` -- but that's for a separate patch!)

tests/test_main.py
asottile-sentry: maybe we should delete the test here and reintroduce it later -- that'd be more clear I think

tests/test_main.py
dbennett-sentry: Yeah, this was a before-times thing.

devenv/lib/config.py
dbennett-sentry: DEFAULT_CONFIG goes away, if that's what you're referring to. Otherwise, yes.

PR #119, opened 2024-04-19
Add module definition; standardize method invocation

dbennett-sentry:


devenv/bootstrap.py
```diff
@@ -13,22 +12,32 @@
...
+from devenv.lib.modules import ExitCode
+
+
+def main(context: Context, argv: Sequence[str] | None = None) -> ExitCode:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-d",
+        "--default-config",
+        action="append",
+        help="Provide a default config value. e.g., -s coderoot:path/to/root",
```
asottile-sentry, line 27: `-s` vs. `-d`

devenv/bootstrap.py
```diff
@@ -13,22 +12,32 @@
...
+    )
 
-help = "Bootstraps the development environment."
-ExitCode: TypeAlias = "str | int | None"
+    args = parser.parse_args(argv)
 
+    configs = {
+        k: v for k, v in [i.split(":") for i in args.default_config or []]
```
asottile-sentry, line 33: probably `.split(1)` so that values can contain `:`

devenv/lib/modules.py
```diff
@@ -0,0 +1,36 @@
...
+    help: str
+    action: Action
+
+
+def require(var: str, message: str) -> Callable[[Action], Action]:
+    def outer(main: Action) -> Action:
+        def inner(context: Context, args: Sequence[str] | None) -> ExitCode:
+            if context.get(var) is None:
+                print(message)
+                return 1
```
asottile-sentry, line 28: this can be `raise SystemExit(message)` since that'll handle unwinding and also conveniently prints the message!

devenv/main.py
```diff
@@ -67,47 +33,48 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
...
+    for name, info in modinfo_list:
+        # Argparse stuff
+        child = subparser.add_parser(info.command, help=info.help)
+        child.add_argument(
+            "--xixax",
+            dest="_module_exec",
+            default=info.action,
+            required=False,
+            help=argparse.SUPPRESS,
         )
```
asottile-sentry, line 62: this is clever -- but very hard to follow -- here's a simpler alternative that doesn't require a secret argument:

```python
modules = {
    mod.name: mod for mod in [...]
}
return modules[args.command].whatever(...)
```

devenv/main.py
```diff
@@ -67,47 +33,48 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
...
+        "repo": Repository(current_root) if current_root else None,
+    }
 
-    if args.command == "doctor":
-        return doctor.main(context, remainder)
-    if args.command == "sync":
-        return sync.main(context, remainder)
-
-    return 1
+    return cast(ExitCode, args._module_exec(context, remainder))
```
asottile-sentry, line 73: `cast` has runtime overhead and generally should only be used for bugs in the type checker -- https://www.notion.so/sentry/sentry-typing-guidelines-best-practices-471d5aacf262476188982ce70f909e00?pvs=4#4c5f66d5e5f2460991cf1074111fdd85

devenv/bootstrap.py
dbennett-sentry: Ack!

devenv/main.py
dbennett-sentry: I think I picked this up from sync(). I'll remove it and check to make sure it still works.

devenv/main.py
```diff
@@ -67,47 +33,49 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
...
-    if args.command == "doctor":
-        return doctor.main(context, remainder)
-    if args.command == "sync":
-        return sync.main(context, remainder)
-
-    return 1
+    command_actions = {info.command: info.action for info in modinfo_list}
+    action = command_actions.get(args.command)
+    assert action is not None
+    return action(context, remainder)
```
IanWoodard, line 74: Played around with this a bit and found it worked quite well. One area that could use improvement is the help messages which give very little if any useful information (for specific actions). This doesn't have to be addressed in this PR, but something to keep an eye on.

IanWoodard (approved):

devenv/doctor.py
```diff
@@ -173,14 +175,11 @@ def main(context: Dict[str, str], argv: Sequence[str] | None = None) -> int:
     )
     args = parser.parse_args(argv)
 
-    match_tags: set[str] = set(args.tag if args.tag else ())
-
     repo = context["repo"]
-    if repo not in {"sentry", "getsentry", "devenv"}:
-        print(f"repo {repo} not supported yet!")
-        return 1
+    assert repo is not None
```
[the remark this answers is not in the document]
dbennett-sentry: This is to inform the type checker that it is not None. If it were None, we'd have problems anyway so I'm using an assert instead of an if.

PR #121, opened 2024-04-22
Decentralized module loading and argument definitions

dbennett-sentry:


devenv/commands/doctor.py
```diff
@@ -232,9 +234,4 @@ def main(context: Context, argv: Sequence[str] | None = None) -> int:
     return 1
 
 
-module_info = DevModuleInfo(
-    action=main,
-    name=__name__,
-    command="doctor",
-    help="Diagnose common issues, and optionally try to fix them.",
-)
+module_info = ModuleDef(module_name=__name__, name="doctor", help="doctor")
```
dbennett-sentry, line 237: Need to update the help on these commands.

asottile-sentry:
2c: this seems "cute" but going to be a pain to debug.  I think we should just be using argparse `subparsers` directly -- it'll be simpler, easier to understand, and much much easier to debug and reason about -- can make simple mutators like [this](https://github.com/pre-commit/pre-commit/blob/85fe18253f6c1959369145257ddc12a374706be3/pre_commit/main.py#L56) for any common setup

or if we want a more involved cli framework (I don't think we do?) we should just use something that exists like click (I don't really like personally) rather than writing our own.  I don't think our subpcommands are complex enough to warrant one though

dbennett-sentry:
> 2c: this seems "cute" but going to be a pain to debug. I think we should just be using argparse `subparsers` directly -- it'll be simpler, easier to understand, and much much easier to debug and reason about -- can make simple mutators like [this](https://github.com/pre-commit/pre-commit/blob/85fe18253f6c1959369145257ddc12a374706be3/pre_commit/main.py#L56) for any common setup


The goal isn't to be cute, it's to reduce friction to every other dev to contribute but to do it in an opinionated way. Argparse isn't the friendliest, and argparse subparsers add new weirdness. 

Let me take a look at your counter-offer, though, as it's stood a bit more of a test of time than what I have tossed together here.

IanWoodard:
> > 2c: this seems "cute" but going to be a pain to debug. I think we should just be using argparse `subparsers` directly -- it'll be simpler, easier to understand, and much much easier to debug and reason about -- can make simple mutators like [this](https://github.com/pre-commit/pre-commit/blob/85fe18253f6c1959369145257ddc12a374706be3/pre_commit/main.py#L56) for any common setup
> 
> The goal isn't to be cute, it's to reduce friction to every other dev to contribute but to do it in an opinionated way. Argparse isn't the friendliest, and argparse subparsers add new weirdness. I wanted a pidgin argument capability, not an all-encompassing alternative.
> 
> Let me take a look at your counter-offer, though, as it's stood a bit more of a test of time than what I have tossed together here.

I appreciate how we’ve managed to keep main.py clean and streamlined the process of adding new arguments without needing to modify multiple files. However, I do see where Anthony is coming from regarding the complexity introduced by our custom parsing layer.

When looking at our _convert_argument_params function and the way we wrap argparse functionalities, I wonder if we might streamline these processes a bit. Simplifying or even reducing some of the custom parsing steps, while maintaining our decorator-based interface, could help keep the system both easy for devs to contribute to and more maintainable overall.

What do you think about refining the interaction between our decorators and argparse subparsers to simplify things? I believe this could help us preserve the developer-friendly nature of our code, while also making it easier to maintain and debug.

dbennett-sentry:
> What do you think about refining the interaction between our decorators and argparse subparsers to simplify things? I believe this could help us preserve the developer-friendly nature of our code, while also making it easier to maintain and debug.

Feel free to give it a go. Requirements are: 

1. extremely low friction for external developers; 
2. opinionated and consistent outcomes; 
3. escape hatches for power-users.

asottile-sentry:
personally I don't think we want low friction for outsiders adding commands to `devenv` -- we want to keep them generic and opinionated

dbennett-sentry:
> personally I don't think we want low friction for outsiders adding commands to `devenv` -- we want to keep them generic and opinionated

That's a difference we'll have to chat through. We'll never get alignment on code if we have differing goals.

PR #123, opened 2024-05-16
Use system python for venv on CI

jan-auer:
Devenv could be useful to setup the venv in CI, though at least in GHA `actions/setup-python` is much faster than a manual download of the matching python version. In CI, we usually take greater care for selecting the right python version already, so devenv could use that when it matches the expected version.

Downside of the current approach is that it silently falls back to downloading a new version if the system provided one doesn't match, which can regress CI times. That should be pretty visible in PRs, though.

### DIFF
--- a/devenv/pythons.py
+++ b/devenv/pythons.py
@@ -1,9 +1,38 @@
 from __future__ import annotations
 
 import os
+import shutil
+import subprocess
 
+from devenv.constants import CI
 from devenv.constants import root
 from devenv.lib import archive
+
+
+def _check_version(exe: str, version: str) -> bool:
+    """Returns true if the executable at the given path is a python interpreter
+    with a semver-compatible version."""
+
+    try:
+        line = (
+            subprocess.check_output([exe, "--version"], text=True)
+            .strip()
+            .lower()
+        )
+    except subprocess.CalledProcessError:
+        return False
+
+    if not line.startswith("python "):
+        return False
+
+    actual = line[7:].split(".")
+    expected = version.split(".")
+
+    return (
+        len(actual) > 2
+        and actual[:2] == expected[:2]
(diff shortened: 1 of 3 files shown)


devenv/pythons.py
```diff
@@ -1,17 +1,51 @@
...
 from devenv.lib import archive
 
 
+def _check_version(exe: str, version: str) -> bool:
+    """Returns true if the executable at the given path is a python interpreter
+    with a semver-compatible version."""
+
+    try:
+        line = (
+            subprocess.check_output([exe, "--version"], text=True)
```
asottile-sentry, line 18: much more reliable to `sys.executable -c 'import json, sys; print(json.dumps(sys.version_info[:3]))'

devenv/pythons.py
```diff
@@ -1,37 +1,22 @@
...
-    actual = line[7:].split(".")
     expected = version.split(".")
 
     return (
-        len(actual) > 2
-        and actual[:2] == expected[:2]
-        and actual[2] >= expected[2]
+        sys.version_info[:2] == expected[:2]
+        and sys.version_info[2] >= expected[2]
     )
```
asottile-sentry, line 20: can just use `sys.version_info >= expected` -- tuple comparisons work out nicely

though you'll have to convert the version to a sequence of integers `expected = [int(part) for part in version.split('.')]`

jan-auer: oh yeah, i didn't run that yet. though i want major + minor to match exactly, just patch needs to be more recent. trying to emulate semver semantics

asottile-sentry: ah yeah then I'd do `sys.version_info[:2] == version[:2] and sys.version_info >= version`

asottile-sentry (approved):

PR #124, opened 2024-06-10
feat: /usr/bin/script wrapper to send full tty output to sentry

joshuarli:
devenv now executes itself with script to tee the full tty output into a log file, which is then uploaded to sentry as an attachent to an event in the case of an unsuccessful exit.

I wasn't able to find a way to upload an attachment to an event after-the-fact though, so we're missing out on any detailed stacktrace that occured in devenv.

TODO: see if we can fetch the event title because these attachment names are not that helpful:

image

### DIFF
--- a/devenv/main.py
+++ b/devenv/main.py
@@ -93,15 +93,17 @@ def main() -> ExitCode:
 
     # the reason we're subprocessing instead of os.execv(cmd[0], cmd)
     # is that script must exit (so that the complete log file is committed to disk)
-    # before sentry sends the event.
+    # before sentry sends the event...
     rc = subprocess.call(cmd)
 
     if rc == 0:
         return rc
 
-    # we're just using sentry-sdk to send the contents of an attachment
-    # i'd love to be able to set it up in the child and retrieve the event id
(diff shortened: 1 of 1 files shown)


devenv/main.py
```diff
@@ -75,17 +77,62 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
...
+
+    # the reason we're subprocessing instead of os.execv(cmd[0], cmd)
+    # is that script must exit (so that the complete log file is committed to disk)
+    # before sentry sends the event...
+    rc = subprocess.call(cmd)
+
+    if rc == 0:
+        return rc
+
+    # i'd love to be able to send a full event in the child then
```
asottile-sentry, line 102: I wonder if we can do something something trace id and have both exist? and just link them after the fact?

joshuarli: ahhh that might work, i'm thinking we start a transaction in the parent and start a span for the child process then only transaction.finish() if an error happens in the child

not sure what the transaction would look like in sentry ui (like, can we easily see the child error event) but gonna give it a go

PR #125, opened 2024-06-11
Add homebrew_repo and homebrew_bin constants for linux

aliu39:
https://docs.brew.sh/Homebrew-on-Linux


devenv/constants.py
```diff
@@ -36,3 +37,6 @@
 if INTEL_MAC:
     homebrew_repo = "/usr/local/Homebrew"
     homebrew_bin = "/usr/local/bin"
+if LINUX:
+    homebrew_repo = "/home/linuxbrew/.linuxbrew"
+    homebrew_bin = f"{homebrew_repo}/bin"
```
asottile-sentry, line 42: this is intentionally omitted.  on linux use your package manager to set up docker instead.  (linuxbrew is ~usually a bad solution on linux)

asottile-sentry (changes requested):

devenv/constants.py
aliu39: Got it. It's just that I tried installing on WSL a good while ago, but IIRC, install.sh will error and end early without these variables.

asottile-sentry: start with a bug report first -- it's likely that there's some gaps there since linux isn't officially supported

aliu39: Got it - I'll actually just close this. Someone said it's highly discouraged to use personal machines for work, so I'm not too keen on installing this anymore..

PR #126, opened 2024-06-17
feat: fetch runs post_fetch if available

joshuarli:
https://github.com/getsentry/sentry/pull/72870#discussion_r1643236206

### DIFF
--- a/ci/devenv-bootstrap.sh
+++ b/ci/devenv-bootstrap.sh
@@ -17,6 +17,7 @@ EOF
 cd "$HOME"
 # note: colima will be used and is necessary for a docker runtime on
 #       macos GitHub runners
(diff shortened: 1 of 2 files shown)


ci/devenv-bootstrap.sh
```diff
@@ -17,8 +17,13 @@ EOF
 cd "$HOME"
 # note: colima will be used and is necessary for a docker runtime on
 #       macos GitHub runners
+export DEVENV_FETCH_BRANCH=devenv-sync-setup-git
```
joshuarli, line 20: will revert this before merging

IanWoodard (approved):

ci/devenv-bootstrap.sh
joshuarli: actually this needs to be here for now, need to release a new devenv version

PR #127, opened 2024-07-09
devenv: small things

joshuarli:
it's the small things in life...

Wanted to get these things in before I cut a new release for https://github.com/getsentry/sentry/pull/72870.

- adds `version/--version`, and `devenv` by itself now prints full usage (instead of pretty much nothing)
- fixes a little bug with gke-gcloud-auth-plugin not on devenv's path
- devenv venvs gitignore themselves!

### DIFF
--- a/devenv/main.py
+++ b/devenv/main.py
@@ -10,6 +10,7 @@ from devenv import fetch
 from devenv import pin_gha
 from devenv import sync
 from devenv.constants import home
+from devenv.constants import user
 from devenv.constants import version
 from devenv.lib.config import read_config
 from devenv.lib.context import Context
@@ -84,6 +85,9 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
 
 
 def main() -> ExitCode:
+    if user == "root":
+        raise SystemExit("You shouldn't be running devenv as root.")
+
     import sys
     import sentry_sdk
 
--- a/devenv/main.py
+++ b/devenv/main.py
@@ -21,10 +21,6 @@ from devenv.lib.repository import Repository
 
 
 def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
-    if len(argv) > 1 and argv[1] in ("version", "--version"):
-        print(f"devenv {version}")
-        return 0
-
     # determine current repo, if applicable
     fake_reporoot = os.getenv("CI_DEVENV_INTEGRATION_FAKE_REPOROOT")
     if fake_reporoot:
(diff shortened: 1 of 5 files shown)


devenv/main.py
```diff
@@ -59,6 +64,10 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
         # Argparse stuff
         subparser.add_parser(info.command, help=info.help)
 
+    if len(argv) == 1:
+        parser.print_help()
+        return 0
+
```
IanWoodard, line 70: Doesn't argparse print out help by default? Is this help message any different?

devenv/main.py
```diff
@@ -19,6 +20,10 @@
 
 
 def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
+    if len(argv) > 1 and argv[1] in ("version", "--version"):
+        print(f"devenv {version}")
+        return 0
+
```
IanWoodard, line 26: As is, this won't show up in help since we aren't making use of argparse. We should make this work more cohesive similar to the rest of the commands, even if it is simpler.

devenv/main.py
joshuarli: No, currently `devenv` alone gives

```
usage: devenv [-h] command ...
devenv: error: the following arguments are required: command
```

devenv/main.py
joshuarli: The problem is, if we have `parser.add_argument("--version", action="store_true")`

then `devenv --version` gives:

```
$ devenv --version
usage: devenv [-h] [--version] command ...
devenv: error: the following arguments are required: command
```

I just gave up quickly and opted for this instead. I think it's fine.

devenv/main.py
IanWoodard: Could we instead utilize the description argument when we construct ArgParse? That way we can utilize argparse's built-in behavior

devenv/main.py
IanWoodard: I think it would be worthwhile to spend the time to figure out how to do this using argparse so we can provide a cohesive cli where every command is listed in the help. If this PR needs to get out quick, I'd suggest leaving this part off and addressing it in a separate PR.

joshuarli: Yeah, i'd like to get this out quick so I can unblock a new release + remaining sentry cleanup. Can polish this stuff later.

devenv/main.py
joshuarli: I don't understand what you mean? Also if this can be further improved upon i'd like to followup later.

devenv/main.py
```diff
@@ -84,6 +85,9 @@ def devenv(argv: Sequence[str], config_path: str) -> ExitCode:
 
 
 def main() -> ExitCode:
+    if user == "root":
```
joshuarli, line 88: this is for https://github.com/getsentry/terraform-sandboxes.private/pull/83#issuecomment-2221498154

IanWoodard (approved):

PR #133, opened 2024-08-09
fix: Use correct .zshrc location when $ZDOTDIR set

keeakita:
Spawn a new shell process to get the value of $ZDOTDIR, if set. This is needed so Homebrew will install correctly for users who have moved their zsh config directory. Note that we cannot simply read this from the environment, as one of the common methods of setting this value (via ~/.zshenv), results in it being a shell variable and not entering the environment.

### DIFF
--- a/devenv/lib/fs.py
+++ b/devenv/lib/fs.py
@@ -1,15 +1,31 @@
 from __future__ import annotations
 
 import os
+import subprocess
 
 from devenv.constants import home
 from devenv.constants import shell
 from devenv.lib import proc
 
 
+def zdotdir() -> str:
+    # Note that we can't simply check os.environ; the most common way of setting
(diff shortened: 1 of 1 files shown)


devenv/lib/fs.py
```diff
@@ -1,15 +1,25 @@
...
+def zdotdir() -> str:
+    return subprocess.run(
+        [shell, "-c", "echo $ZDOTDIR"], text=True, capture_output=True
+    ).stdout.strip()
+
+
 def shellrc() -> str:
     if shell == "zsh":
-        return f"{home}/.zshrc"
+        dotdir = zdotdir()
```
joshuarli, line 19: I'd check for `~/.zshrc` to exist first so we don't have an extra subprocess for the common case.

Could you also add your PR description as a comment? It's not obvious.

joshuarli (approved):

PR #134, opened 2024-08-21
feat: ensure rosetta before colima start

joshuarli:
on a new mac arm dev machine, if rosetta isn't installed, colima start --vz-rosetta will just warn about it and blissfully proceed without rosetta

it's pretty cheap to ensure rosetta's installed + running so let's just unconditionally do this before starting colima

### DIFF
--- a/devenv/lib/rosetta.py
+++ b/devenv/lib/rosetta.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import os
+
+from devenv.constants import DARWIN
+from devenv.constants import MACHINE
+from devenv.lib import proc
+
+
+def ensure() -> None:
+    if not (DARWIN and (MACHINE == "arm64")):
+        return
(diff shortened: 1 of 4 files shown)


devenv/lib/rosetta.py
```diff
@@ -0,0 +1,31 @@
...
+from devenv.constants import DARWIN
+from devenv.constants import MACHINE
+from devenv.lib import proc
+
+
+def ensure() -> None:
+    if not (DARWIN and (MACHINE == "arm64")):
+        return
+
+    # this file doesn't exist if rosetta 2 isn't installed
```
joshuarli, line 14: actually just provisioning a mac arm UTM VM as we speak to verify my guess

(it's really hard to uninstall rosetta after it's installed)

asottile-sentry (approved):

PR #137, opened 2024-08-27
feat: download retries

joshuarli:
also adds a unittest

### DIFF
--- a/devenv/lib/archive.py
+++ b/devenv/lib/archive.py
@@ -28,7 +28,8 @@ def download(
     retries: int = 3,
     retry_exp: float = 2.0,
 ) -> str:
-    assert retries >= 0
+    if retries < 0:
+        raise ValueError("Retries cannot be negative")
 
     if not dest:
         cache_root = f"{home}/.cache/sentry-devenv"
--- a/devenv/lib/archive.py
+++ b/devenv/lib/archive.py
@@ -6,6 +6,7 @@ import secrets
 import shutil
 import tarfile
 import tempfile
+import time
 import urllib.request
 from urllib.error import HTTPError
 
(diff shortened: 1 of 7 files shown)


devenv/lib/archive.py
```diff
@@ -20,17 +21,34 @@ def atomic_replace(src: str, dest: str) -> None:
...
 
-def download(url: str, sha256: str, dest: str = "") -> str:
+def download(
+    url: str,
+    sha256: str,
+    dest: str = "",
+    retries: int = 3,
+    retry_exp: float = 2.0,
+) -> str:
+    assert retries >= 0
```
IanWoodard, line 31:
```suggestion
    if retries < 0:
        raise ValueError("Retries cannot be negative")
```
Since assertions can be disabled, I feel an explicit check would be better

devenv/lib/archive.py
```diff
@@ -20,17 +21,34 @@ def atomic_replace(src: str, dest: str) -> None:
...
         dest = f"{cache_root}/{sha256}"
         os.makedirs(cache_root, exist_ok=True)
 
     if not os.path.exists(dest):
-        try:
-            resp = urllib.request.urlopen(url)
-        except HTTPError as e:
-            raise RuntimeError(f"Error getting {url}: {e}")
+        retry_sleep = 1.0
+        while retries >= 0:
```
IanWoodard, line 40:
```suggestion
        while retries > 0:
```

joshuarli: why? currently if retries = 0 then we'll try once and if err, raise the runtimeerror

devenv/lib/archive.py
joshuarli: good one!

joshuarli:
ci needs https://github.com/getsentry/sentry/pull/76688

devenv/lib/archive.py
IanWoodard: Fair enough

IanWoodard (approved):

PR #138, opened 2024-08-28
Add prerequisites

natemoo-re:
Adding a prerequisites section so that future travelers aren't as confused as I was why `devenv bootstrap` fails.

Users running that command will need to be part of the `getsentry/engineering` team or set a `SENTRY_EXTERNAL_CONTRIBUTOR` environment variable. This should be mentioned up front!


IanWoodard (approved):

joshuarli (approved):
Thanks!

PR #139, opened 2024-08-29
feat: verbose

joshuarli:


devenv/sync.py
```diff
@@ -11,6 +12,13 @@
...
 def main(context: Context, argv: Sequence[str] | None = None) -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-v", "--verbose", action="store_true")
+    args = parser.parse_args(argv)
+
+    # also support devenv sync -v in addition to devenv -v sync
+    verbose = context["verbose"] or args.verbose
+
     repo = context["repo"]
     assert repo is not None
```
joshuarli, line 23: hmm i'm trying to use typeguard so i can remove this assert but it's getting a little messy... https://github.com/getsentry/devenv/commit/3cf750971e4637400dc44aea4cb40cab4ed858be

PR #141, opened 2024-09-18
allow .devenv/node to error when checking health

asottile-sentry:


asottile-sentry:
best viewed with [?w=1](https://github.com/getsentry/devenv/pull/141/files)

joshuarli (approved):

PR #142, opened 2024-09-18
properly quote shell variables

asottile-sentry:


asottile-sentry:
@joshuarli we're going to need this in a lot of other places lol

asottile-sentry:
> @joshuarli we're going to need this in a lot of other places lol

also `/usr/bin/env` shouldn't be needed I think

saponifi3d (approved):
+1 - thanks for fixing!

(just a heads up, I noticed this in yarn, npm, pnpm, and npx as well)

PR #143, opened 2024-09-18
quote all the shims and remove unncessary /usr/bin/env

joshuarli:
f/u to https://github.com/getsentry/devenv/pull/142

properly quotes everything in shims and removes unnecessary /usr/bin/env!

### DIFF
--- a/devenv/lib/gcloud.py
+++ b/devenv/lib/gcloud.py
@@ -26,19 +26,21 @@ def _install(url: str, sha256: str, into: str) -> None:
     # and we may as well reuse devenv's internal python.
     fs.write_script(
         f"{into}/gcloud",
-        f"""#!/bin/sh
+        """#!/bin/sh
 export CLOUDSDK_PYTHON="{root}/python/bin/python3" \
        PATH="{into}/google-cloud-sdk/bin:${{PATH}}"
 exec gcloud "$@"
 """,
+        shell_escape={"root": root, "into": into},
     )
     fs.write_script(
         f"{into}/gsutil",
-        f"""#!/bin/sh
+        """#!/bin/sh
 export CLOUDSDK_PYTHON="{root}/python/bin/python3" \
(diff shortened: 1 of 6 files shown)


devenv/lib/node.py
```diff
@@ -126,7 +127,8 @@ def install_yarn(version: str, reporoot: str) -> None:
     fs.write_script(
         f"{binroot}/yarn",
         f"""#!/bin/sh
-exec /usr/bin/env PATH={binroot}/node-env/bin:"$PATH" {binroot}/node-env/bin/yarn "$@"
+export PATH="{binroot}/node-env/bin:${{PATH}}"
```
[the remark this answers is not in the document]
joshuarli: `exec: FOO=1: not found`

```
#!/bin/sh
exec FOO=1 bash -c 'env | grep FOO'
```

devenv/lib/gcloud.py
```diff
@@ -27,13 +27,17 @@ def _install(url: str, sha256: str, into: str) -> None:
     fs.write_script(
         f"{into}/gcloud",
         f"""#!/bin/sh
-exec /usr/bin/env CLOUDSDK_PYTHON={root}/python/bin/python3 PATH={into}/google-cloud-sdk/bin:$PATH gcloud "$@"
+export CLOUDSDK_PYTHON="{root}/python/bin/python3" \
+       PATH="{into}/google-cloud-sdk/bin:${{PATH}}"
```
asottile-sentry, line 31: technically we should be using `shlex.quote` on all of these values (protect against someone `mkdir '$(cat /etc/passwd)'`)

joshuarli: should i make write_script better by passing an optional format dict that'll then shell escape all those tokens?

asottile-sentry: if it makes it simpler sure that sounds decent

asottile-sentry (approved):

PR #147, opened 2024-09-24
feat: devenv delete

joshuarli:
Closes https://github.com/getsentry/devenv/issues/89.

This provides an easy way to blow away a repo's dev environment if needed.

Also included which may be useful is deleting global colima state + uninstalling devenv itself.


asottile-sentry:
I am not a fan of advertising something like this -- people are going to waste lots of time with this and not report actual fixable problems

PR #148, opened 2024-09-27
ref: silence first time ssh access check error during bootstrap

joshuarli:
bootstrap on a new machine will have this non-error printed which is confusing - silence it the first time because ssh keys aren't provisioned yet

```
Command `sh -c 'ssh -T git@github.com 2>&1'` failed! (code 255)
combined out:
git@github.com: Permission denied (publickey).

Are you a Sentry employee? (Y/n): 
```


bc-sentry (approved):
LGTM thanks for fixing so fast.

PR #149, opened 2024-09-30
fix: chdir to reporoot before syncing

joshuarli:
for example in sentry if you're in `src/sentry` then 

```
$ devenv sync                                                                                                 0
ensuring sentry venv at .venv...
virtualenv doesn't exist or is using an outdated python, recreating at .venv...
Could not import runpy module
Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1176, in exec_module
  File "<frozen runpy>", line 15, in <module>
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1176, in exec_module
  File "<frozen importlib.util>", line 168, in <module>
AttributeError: module 'types' has no attribute 'ModuleType'
Command `/Users/josh/.local/share/sentry-devenv/pythons/3.12.6/python/bin/python3 -m venv .venv` failed! (code 1)
```

```
ensuring sentry venv at .venv...
> /Users/josh/dev/devenv/devenv/lib/venv.py(107)check()
-> if not os.path.exists(f"{venv}/pyvenv.cfg"):
(Pdb) p os.getcwd()
'/Users/josh/dev/sentry/src/sentry'
```


devenv/sync.py
```diff
@@ -30,6 +30,8 @@ def main(context: Context, argv: Sequence[str] | None = None) -> int:
         "repo": repo.name,
         "coderoot": context.get("code_root"),
     }
+
+    os.chdir(repo.path)
     return module.main(context_compat)  # type: ignore
```
asottile-sentry, line 35: I would:

```python
with contextlib.chdir(repo.path):
    return module.main(...)
```

that way it is properly undone if a test calls this (should we add a test to demo this behaviour?)

joshuarli: oooh that's good

devenv/sync.py
```diff
@@ -30,7 +31,9 @@ def main(context: Context, argv: Sequence[str] | None = None) -> int:
         "repo": repo.name,
         "coderoot": context.get("code_root"),
     }
-    return module.main(context_compat)  # type: ignore
+
+    with contextlib.chdir(repo.path):
```
joshuarli, line 35: @asottile-sentry can i drop 3.10 from tox? not sure why you wanted to keep it around, the only comment says "for internal pypi compat"

asottile-sentry: it looks like ops:salt still uses python 3.10 otherwise we could drop it from pypi I think

joshuarli: sync.py is run with devenv's python 3.11 (see install-devenv.sh), that 3.10 venv is only used for salt

asottile-sentry (approved):

Issue #146, opened 2024-09-24
colima and lima upgrade


joshuarli:
https://github.com/abiosoft/colima/issues/949#issuecomment-2317569867

the double-start issue has since been fixed - we had a workaround for this before, but it would be good to upgrade past that

it must be noted that lima >=0.23.2 is necessary (see https://github.com/lima-vm/lima/issues/2556)

joshuarli:
getting this with limactl 0.23.2, with both colima versions. Not an issue with 0.19.1 (what we're currently on). 
related https://github.com/lima-vm/lima/issues/1182

```
> Starting the instance "colima" with VM driver "vz"
> "Attempting to download the image" arch=aarch64 digest= location=/Users/josh/.cache/colima/caches/fc0bfb9256e1f61294f2c41fc95ff4feb6c0daafccfb3b435e2d972e39bc8432.raw
> Downloaded the image from "/Users/josh/.cache/colima/caches/fc0bfb9256e1f61294f2c41fc95ff4feb6c0daafccfb3b435e2d972e39bc8432.raw"
> Converting "/Users/josh/.colima/_lima/colima/basedisk" (raw) to a raw disk "/Users/josh/.colima/_lima/colima/diffdisk"
> Expanding to 60GiB
> [hostagent] hostagent socket created at /Users/josh/.colima/_lima/colima/ha.sock
> [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/josh/.colima/_lima/colima/serial*.log")
> [hostagent] Setting up Rosetta share
> [hostagent] Unable to configure Rosetta: failed to create a new rosetta directory share caching option: unsupported build target macOS version for 14.0 (the binary was built with __MAC_OS_X_VERSION_MAX_ALLOWED=130100; needs recompilation)
> SSH Local Port: 54364
> [hostagent] Waiting for the essential requirement 1 of 2: "ssh"
> [hostagent] [VZ] - vm state change: running
> [hostagent] Waiting for the essential requirement 1 of 2: "ssh"
> [hostagent] The essential requirement 1 of 2 is satisfied
> [hostagent] Waiting for the essential requirement 2 of 2: "user session is ready for ssh"
> [hostagent] The essential requirement 2 of 2 is satisfied
> [hostagent] Waiting for the guest agent to be running
> [hostagent] Forwarding "/var/run/docker.sock" (guest) to "/Users/josh/.colima/default/docker.sock" (host)
[...]

PR #152, opened 2024-10-02
feat: new stripN and ability to download blobs from homebrew

joshuarli:
tl;dr we need to upgrade to limactl 0.23.2 and only the homebrew builds work since their github releases aren't compiled for macos 14+, and the vast majority of limactl consumers get it from homebrew

https://github.com/getsentry/devenv/issues/146#issuecomment-2384323691

their homebrew releases have a different structure - a `lima/0.23.2` prefix - so we do need a stripN after all

i've also improved the implementation, it now determines the prefix to be stripped one time then expects to strip that same prefix from all members - otherwise it's an unexpected structure

### DIFF
--- a/devenv/lib/archive.py
+++ b/devenv/lib/archive.py
@@ -87,26 +87,63 @@ def download(
 
 
 # mutates members!
-# strips the leading component, and asserts that
-# the same component was stripped across all members
+# strips N leading components and optionally adds a new prefix
+# (what ends up being stripped should be a common prefix for all entries)
 # (/ is always stripped and doesn't count)
-def strip1(members: Sequence[tarfile.TarInfo]) -> None:
-    for member in members:
-        i = member.path.find("/")
-        if i == -1:
+def strip_n(
+    members: Sequence[tarfile.TarInfo], strip_n: int, new_prefix: str = ""
+) -> None:
+    # we'll use the first member to determine the prefix to strip
+    member = members[0]
+
+    end = 0
+    if member.path.find("/") == 0:
+        # strip leading "/"
+        end = 1
+
+    for n in range(strip_n):
+        next_at = member.path[end:].find("/")
+        if next_at == -1 and n != strip_n - 1:
+            # no more '/' but we're not done iterating
+            # this means this member isn't nested as deep as
+            # N directories we want to strip, which is
+            # unexpected
             raise ValueError(
-                f"unexpected archive structure: no component left to strip in {member.path}"
+                f"""unexpected archive structure:
+
+trying to strip {strip_n} leading components but {member.path} isn't that deep
+"""
             )
-        elif i == 0:
-            i = member.path[1:].find("/") + 1
-            if i == 0:
-                raise ValueError(
-                    f"unexpected archive structure: no component left to strip in {member.path}"
-                )
+        end += next_at + 1
 
-        member.path = member.path[i + 1 :]  # noqa: E203
+    if end == 0 and strip_n > 0:
+        # special case where members are just top level files
+        raise ValueError(
+            f"""unexpected archive structure:
 
-    # TODO: assert that the same component was stripped across all members
+trying to strip {strip_n} leading components but {member.path} isn't that deep
+"""
+        )
+
+    stripped_prefix = member.path[:end]
+
+    for member in members:
+        if not member.path.startswith(stripped_prefix):
+            raise ValueError(
+                f"""unexpected archive structure:
+
+{member.path} doesn't have the prefix to be removed ({stripped_prefix})
+"""
+            )
+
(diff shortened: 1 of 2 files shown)


tests/lib/test_archive.py
```diff
@@ -121,6 +153,19 @@ def test_unpack_tar(tar: pathlib.Path, tmp_path: pathlib.Path) -> None:
     assert dest.joinpath("hello.txt").read_text() == "hello world\n"
 
 
+def test_unpack_tar_strip1(tar: pathlib.Path, tmp_path: pathlib.Path) -> None:
+    dest = tmp_path.joinpath("dest")
+    archive.unpack(
```
joshuarli, line 158: erm this should be raising since there's no component to strip, but i need to go eat lunch soon

devenv/lib/archive.py
```diff
@@ -77,22 +87,54 @@ def download(
...
+        end = 1
+
+    for n in range(strip_n):
+        next_at = member.path[end:].find("/")
+        if next_at == -1 and n != strip_n - 1:
+            # no more '/' but we're not done iterating
+            # this means this member isn't nested as deep as
+            # N directories we want to strip, which is
+            # unexpected
+            raise RuntimeError(
```
asottile-sentry, line 111: ValueError is probably a better exception

devenv/lib/archive.py
```diff
@@ -77,22 +87,54 @@ def download(
 
 
 # mutates members!
-# strips the leading component (/ is always stripped and doesn't count)
-# and optionally replaces with a new prefix
-def strip1(members: Sequence[tarfile.TarInfo], new_prefix: str = "") -> None:
+# strips N leading components and optionally adds a new prefix
+# (what ends up being stripped should be a common prefix for all entries)
+# (/ is always stripped and doesn't count)
+def stripN(
```
asottile-sentry, line 93: probably would call this `strip_n`

devenv/lib/archive.py
```diff
@@ -77,22 +87,63 @@ def download(
 
 
 # mutates members!
-# strips the leading component (/ is always stripped and doesn't count)
-# and optionally replaces with a new prefix
-def strip1(members: Sequence[tarfile.TarInfo], new_prefix: str = "") -> None:
+# strips N leading components and optionally adds a new prefix
+# (what ends up being stripped should be a common prefix for all entries)
+# (/ is always stripped and doesn't count)
+def strip_n(
```
asottile-sentry, line 93: without rewriting `strip1` -- can't `strip_n` just be a composition of `n` `strip1`s?

joshuarli: i guess it can, but i'm so proud of my unnecessary optimization

asottile-sentry (approved):

asottile-sentry (approved):

joshuarli:
ah we've regressed on a real world case caught by ci, i should have used this as a test case to begin with

`trying to strip 1 leading components but node-v20.13.1-darwin-arm64 isn't that deep`

```
node-v20.13.1-linux-x64/
node-v20.13.1-linux-x64/share/
node-v20.13.1-linux-x64/share/doc/
```

tests/lib/test_archive.py
```diff
@@ -44,6 +44,87 @@ def tgz(tmp_path: pathlib.Path) -> pathlib.Path:
...
+def tar5(tmp_path: pathlib.Path) -> pathlib.Path:
+    a = tmp_path.joinpath("a")
+    a.write_text("")
+    b = tmp_path.joinpath("b")
+    b.write_text("")
+
+    tar = tmp_path.joinpath("tar")
+
+    with tarfile.open(tar, "w:tar") as tarf:
+        tarf.add(a, arcname="foo/v1/bar")
```
joshuarli, line 102: so after writing `tar6` i now realize this isn't a valid tarfile, but would it be reasonable to forgo a bunch more proper setup or should i just bite the bullet

asottile-sentry: probably best to set it up properly so it represents a real case (and catches the problme found)

tests/lib/test_archive.py
```diff
@@ -269,7 +273,7 @@ def test_unpack_strip_n_unexpected_structure_inconsistent_components(
         f"{excinfo.value}"
         == """unexpected archive structure:
 
-foo/v2/baz doesn't have the prefix to be removed (foo/v1/)
+trying to strip 2 leading components but foo isn't that deep
```
joshuarli, line 276: ugh so now that i have a properly formed tarfile it reveals a problem with the implementation

if you try and strip 2, you can't use the first member (`foo/`) to determine stripped_prefix, it'll error like this

PR #153, opened 2024-10-02
fix: restore py310 metadata

joshuarli:
partial revert of https://github.com/getsentry/devenv/pull/151


asottile-sentry (approved):
the reason for this is pypi currently requires a 3.10 minimum (until salt is gone) and so this needs to set that minimum (even though some functionality will not work in 3.10) -- it's a calculated risk we're taking and ~sorta lying about the metadata here

PR #154, opened 2024-10-03
ref: colima shim

joshuarli:
this adds a shim for colima that:

1. <s>if start, reexecs as `devenv colima start` - some people in the past have run `colima delete` then `colima start` to try and reset state, however `colima start` has bad defaults and (currently) suffers from double-start problem which `devenv colima start` solves</s>

2. proactively ensures that COLIMA_HOME is ~/.colima. The version I'm trying to upgrade us to prioritizes using XDG_CONFIG_HOME (https://github.com/abiosoft/colima/commit/05fd57766899b5b9c3eed14e8889083a546e3353) if set.

### DIFF
--- a/devenv/lib/colima.py
+++ b/devenv/lib/colima.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import os
 import platform
 import shutil
-import sys
 import tempfile
 from enum import Enum
 
@@ -49,18 +48,9 @@ export COLIMA_HOME="{home}/.colima"
 # needed to ensure COLIMA_HOME is what we want it to be
 unset XDG_CONFIG_HOME
 
-# no one is going to type 'devenv colima start'
-# but it's very important that colima is started with the
-# right parameters
-for arg in "$@"; do
-  if [ "$arg" = "start" ]; then
-    exec "{py}" -P -m devenv colima start
-  fi
-done
-
 exec "{binroot}/colima-bin" "$@"
 """,
-        shell_escape={"binroot": binroot, "home": home, "py": sys.executable},
+        shell_escape={"binroot": binroot, "home": home},
     )
 
 
--- a/devenv/lib/colima.py
+++ b/devenv/lib/colima.py
@@ -5,10 +5,8 @@ import platform
 import shutil
 import tempfile
 from enum import Enum
-from typing import Optional
 
 from devenv.constants import home
-from devenv.constants import root
 from devenv.lib import archive
 from devenv.lib import docker
 from devenv.lib import fs
@@ -23,8 +21,8 @@ def _install(url: str, sha256: str, into: str) -> None:
     os.makedirs(into, exist_ok=True)
     with tempfile.TemporaryDirectory(dir=into) as tmpd:
         archive.download(url, sha256, dest=f"{tmpd}/colima")
-        os.replace(f"{tmpd}/colima", f"{into}/colima")
-        os.chmod(f"{into}/colima", 0o775)
+        os.replace(f"{tmpd}/colima", f"{into}/colima-bin")
+        os.chmod(f"{into}/colima-bin", 0o775)
 
 
 def uninstall(binroot: str) -> None:
@@ -41,15 +39,8 @@ def uninstall(binroot: str) -> None:
             pass
 
 
-def install(
-    version: str, url: str, sha256: str, reporoot: Optional[str] = ""
-) -> None:
-    if reporoot:
-        binroot = fs.ensure_binroot(reporoot)
-    else:
-        # compatibility with devenv <= 1.4.0
-        binroot = f"{root}/bin"
-        os.makedirs(binroot, exist_ok=True)
+def install(version: str, url: str, sha256: str, reporoot: str) -> None:
+    binroot = fs.ensure_binroot(reporoot)
 
     if shutil.which("colima", path=binroot) == f"{binroot}/colima":
         stdout = proc.run((f"{binroot}/colima", "--version"), stdout=True)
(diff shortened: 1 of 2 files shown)


devenv/lib/colima.py
```diff
@@ -41,17 +40,42 @@ def uninstall(binroot: str) -> None:
...
+unset XDG_CONFIG_HOME
+
+# no one is going to type 'devenv colima start'
+# but it's very important that colima is started with the
+# right parameters
+for arg in "$@"; do
+  if [ "$arg" = "start" ]; then
+    exec "{py}" -P -m devenv colima start
+  fi
+done
```
asottile-sentry, line 59: this seems (1) kinda surprising and (2) pretty fragile (what if a random unrelated command also has `start` in it?)

I wouldn't expect `colima` to do anything but actual colima personally -- I realize this is a sharp edge but I think this type of hack is more likely to bite us than it is to help us -- instead maybe we should work on detecting the broken case and healing it?

joshuarli: i'll walk this part back, but i have yet to find a way to change the vm's parameters like file mounts after first-start - will look again into if that's possible then

asottile-sentry (approved):

asottile-sentry:
> 2\. proactively ensures that COLIMA_HOME is ~/.colima. The version I'm trying to upgrade us to prioritizes using XDG_CONFIG_HOME

another thought... what if we went the other way on this and moved everyone to the new location? (would mean we wouldn't need this shim forever)

joshuarli:
> > 2. proactively ensures that COLIMA_HOME is ~/.colima. The version I'm trying to upgrade us to prioritizes using XDG_CONFIG_HOME
> 
> another thought... what if we went the other way on this and moved everyone to the new location? (would mean we wouldn't need this shim forever)

oh, actually COLIMA_HOME isn't even a thing in 0.6.6 (added in colima@fd807660495bd29df573383f030e60d580ea07e5)

in 0.6.6 if you have XDG_CONFIG_HOME set it'll still use ~/.colima because macos (lol), so there's no way to not use ~/.colima

```
			dir = filepath.Join(dir, ".colima")
			_, err = os.Stat(dir)
			// TODO: remove macOS when QEMU_SYSTEM_ARCH is handled properly upstream.
			if err == nil || util.MacOS() {
				return dir, nil
			}
			// else
			dir = os.Getenv("XDG_CONFIG_HOME")
			if dir != "" {
				return filepath.Join(dir, "colima"), nil
			}
			// else
			dir, err = os.UserConfigDir()
			if err != nil {
				return "", err
			}
			return filepath.Join(dir, "colima"), nil
```

I think it's best to keep using `~/.colima`, it's what people are used to and switching over (and potentially reverting) involves needing to totally recreate the vm

devenv/lib/colima.py
```diff
@@ -41,17 +39,29 @@ def uninstall(binroot: str) -> None:
...
+        f"{binroot}/colima",
+        """#!/bin/sh
+export COLIMA_HOME="{home}/.colima"
+
+# needed to ensure COLIMA_HOME is what we want it to be
+unset XDG_CONFIG_HOME
+
+exec "{binroot}/colima-bin" "$@"
+""",
+        shell_escape={"binroot": binroot, "home": home},
```
asottile-sentry, line 53: this is going to double shell escape I think

asottile-sentry: same for `home`

joshuarli:
```sh
#!/bin/sh
export COLIMA_HOME="/Users/josh/.colima"

# needed to ensure COLIMA_HOME is what we want it to be
unset XDG_CONFIG_HOME

exec "/Users/josh/dev/sentry/.devenv/bin/colima-bin" "$@"
```

asottile-sentry: have a space in `home` or `binroot`

joshuarli: hmm

```
fs.write_script(
        "test.sh",
        """#!/bin/sh
export fp="{dir}/foo"
touch "$fp"
""",
        shell_escape={"dir": "/Users/josh/s p a c e"},
    )
```

```
bash test.sh
touch: '/Users/josh/s p a c e'/foo: No such file or directory
```

would the expectation be

`export fp='/Users/josh/s p a c e/foo'`?

joshuarli: ah

```
fs.write_script(
        "test.sh",
        """export fp="{dir}/foo"
mkdir -p "{dir}"
touch "$fp"
test -e "$fp" || echo fail
""",
        shell_escape={"dir": "/Users/josh/s p a c e"},
    )
```

this generates

```
export fp="'/Users/josh/s p a c e'/foo"
mkdir -p "'/Users/josh/s p a c e'"
touch "$fp"
test -e "$fp" || echo fail
```

which doesn't fail but it makes this dir haha...

```
josh@CTKP7QRX4L.local ~  $ tree \'                                                                                                                        0
'
└── Users
    └── josh
        └── s p a c e'
            └── foo
```

joshuarli: need to unquote it seems

```
fs.write_script(
        "test.sh",
        """export fp={dir}/foo
mkdir -p {dir}
touch "$fp"
test -e "$fp" || echo fail
""",
        shell_escape={"dir": "/Users/josh/s p a c e"},
    )
```

...

```
josh@CTKP7QRX4L.local ~  $ tree s\ p\ a\ c\ e/                                                                                                            
s p a c e/
└── foo
```

joshuarli: will apply this to other shims in a followup pr

asottile-sentry (approved):

PR #158, opened 2024-10-08
fix: unquote shell escaped parts of shims but keep $PATH quoted

joshuarli:
f/u to https://github.com/getsentry/devenv/pull/154/commits/8ee991f624ba08115214ed6f793ccb7a3b7cfd83 for the remaining shims

additionally, quotes are now needed around `${{PATH}}` so it becomes `"${PATH}"` after str.format

### DIFF
--- a/devenv/lib/gcloud.py
+++ b/devenv/lib/gcloud.py
@@ -27,8 +27,8 @@ def _install(url: str, sha256: str, into: str) -> None:
     fs.write_script(
         f"{into}/gcloud",
         """#!/bin/sh
-export CLOUDSDK_PYTHON="{root}/python/bin/python3" \
-       PATH="{into}/google-cloud-sdk/bin:${{PATH}}"
+export CLOUDSDK_PYTHON={root}/python/bin/python3 \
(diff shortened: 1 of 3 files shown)


devenv/lib/gcloud.py
```diff
@@ -27,17 +27,17 @@ def _install(url: str, sha256: str, into: str) -> None:
     fs.write_script(
         f"{into}/gcloud",
         """#!/bin/sh
-export CLOUDSDK_PYTHON="{root}/python/bin/python3" \
-       PATH="{into}/google-cloud-sdk/bin:${{PATH}}"
+export CLOUDSDK_PYTHON={root}/python/bin/python3 \
+       PATH={into}/google-cloud-sdk/bin:\"${{PATH}}\"
```
asottile-sentry, line 31: you don't need backslashes inside a triple quoted string

joshuarli: oh nice!

asottile-sentry (approved):

asottile-sentry (approved):

PR #163, opened 2024-10-10
feat: require minimum devenv version for a repo

joshuarli:
sync now checks for an optionally defined minimum devenv version and recommends how to update the global devenv accordingly in order to use it to run sync to completion

### DIFF
--- a/devenv/lib/repository.py
+++ b/devenv/lib/repository.py
@@ -42,10 +42,10 @@ class Repository:
         except KeyError:
             return
 
-        min_major, min_minor, min_patch = map(int, minimum_version.split("."))
-        major, minor, patch = map(int, version.split("."))
+        parsed_version = tuple(map(int, version.split(".")))
+        parsed_minimum_version = tuple(map(int, minimum_version.split(".")))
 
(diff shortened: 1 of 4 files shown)


devenv/lib/repository.py
```diff
@@ -30,3 +32,29 @@ def ensure_binroot(self) -> None:
...
+        cfg = self.config()
+        try:
+            minimum_version = cfg["devenv"]["minimum_version"]
+        except KeyError:
+            return
+
+        min_major, min_minor, min_patch = map(int, minimum_version.split("."))
+        major, minor, patch = map(int, version.split("."))
+
+        if (major, minor, patch) < (min_major, min_minor, min_patch):
```
asottile-sentry, line 48: you don't really need to split these up fwiw -- can just be `parsed_minimum_version` and `parsed_version`

asottile-sentry (approved):
fixnship

devenv/lib/repository.py
joshuarli: ah it's vestigial from when i was thinking about how to compare, before i realized you can just use tuples

PR #164, opened 2024-10-11
feat: unnamed venv defaults to .venv

joshuarli:
`.venv` is a pretty standard pattern and we should explicitly support this and also provide an example in the docs.


joshuarli:
superseded by https://github.com/getsentry/devenv/pull/165 as it's already possible to have a single (named) venv at .venv

PR #165, opened 2024-10-11
doc: add single .venv example

joshuarli:


bukzor (approved):

bukzor:
This is an improvement, but I feel like "a single venv" should be treated as the default case. It feels like a special case, currently.

PR #170, opened 2024-11-18
feat: global colima install

joshuarli:
https://getsentry.atlassian.net/browse/DEVINFRA-434

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -18,31 +18,31 @@ env:
   DEBUG: "1"
 
 jobs:
-  # bootstrap:
-  #   runs-on: ubuntu-22.04
-  #   timeout-minutes: 60
-  #   env:
-  #     DEVENV_FETCH_BRANCH: master
-  #     SNTY_DEVENV_BRANCH:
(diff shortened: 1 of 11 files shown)


joshuarli:
hm so strip isnt working on the lima linux x86-64 archive 

```
tarinfo = <TarInfo 'share/lima/templates/almalinux.yaml' at 0x7fdba9372e00>

    def _find_link_target(self, tarinfo):
        """Find the target member of a symlink or hardlink member in the
           archive.
        """
        if tarinfo.issym():
            # Always search the entire archive.
            linkname = "/".join(filter(None, (os.path.dirname(tarinfo.name), tarinfo.linkname)))
            limit = None
        else:
            # Search the archive before the link, because a hard link is
            # just a reference to an already archived file.
            linkname = tarinfo.linkname
            limit = tarinfo
    
        member = self._getmember(linkname, tarinfo=limit, normalize=True)
        if member is None:
>           raise KeyError("linkname %r not found" % linkname)
E           KeyError: "linkname 'lima/0.23.2/share/lima/templates/almalinux-9.yaml' not found"
```

asottile-sentry (approved):
I feel like this change should be three separate PRs since it's kind of a lot all at once and I've maybe missed some stuff in reviewing it

PR #174, opened 2024-11-25
fix: docker version detection

joshuarli:
forgot to check this in, sorry

will be releasing 1.14.1

### DIFF
--- a/devenv/lib/docker.py
+++ b/devenv/lib/docker.py
@@ -91,7 +91,7 @@ def install_global() -> None:
 
     if shutil.which("docker", path=binroot) == f"{binroot}/docker":
         stdout = proc.run((f"{binroot}/docker", "--version"), stdout=True)
-        installed_version = stdout.strip().split()[2]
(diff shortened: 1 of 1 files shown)


devenv/lib/docker.py
```diff
@@ -91,7 +91,7 @@ def install_global() -> None:
 
     if shutil.which("docker", path=binroot) == f"{binroot}/docker":
         stdout = proc.run((f"{binroot}/docker", "--version"), stdout=True)
-        installed_version = stdout.strip().split()[2]
+        installed_version = stdout.strip().split()[2][:-1]
```
IanWoodard, line 94: Could we use regex instead? Or make clearer that the `[:-1]` is needed to remove the trailing comma (i.e. using `split(',')[0]`)?

joshuarli: you can just `docker --version` and have a look yourself - yes, it's to remove the trailing comma

IanWoodard (approved):

PR #184, opened 2025-01-21
fix(ci): unbound variable error during bootstrap

joshuarli:
since the python 3.13 upgrade the bootstrap job has been failing:

```
failed command (code 2):
    make apply-migrations

Output:
/home/runner/code/sentry/scripts/lib.sh: line 67: red: unbound variable
make: *** [Makefile:23: apply-migrations] Error 1
```

this is because in sentry:

```
$ CI=1 make apply-migrations                                                                                        2
/Users/josh/dev/sentry/scripts/lib.sh: line 67: red: unbound variable
make: *** [apply-migrations] Error 1
```

updating our devenv config will avoid this as we pass the ifs in sentry's `query-valid-python-version` 

will f/u with a real fix in sentry (@asottile-sentry should we just remove the `query-valid-python-version` piece altogether?)

### DIFF
--- a/ci/devenv-bootstrap.sh
+++ b/ci/devenv-bootstrap.sh
@@ -54,21 +54,23 @@ export VIRTUAL_ENV="${HOME}/code/sentry/.venv"
 # so we don't break when upstream config is updated
 cat <<EOF > "devenv/config.ini"
 [venv.sentry]
-python = 3.12.3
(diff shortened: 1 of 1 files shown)


asottile-sentry (approved):
yeah the red thing has been broken for a while -- it's only conditionally defined and then used unconditionally

PR #187, opened 2025-04-04
feat: install docker-buildx

joshuarli:
The global docker install doesn't come with buildx. We should install it - some local development expects buildx to be available. I was building gocd images and they use buildx.

### DIFF
--- a/devenv/lib/docker.py
+++ b/devenv/lib/docker.py
@@ -100,7 +100,13 @@ def _check_buildx(binroot: str, expected_version: str) -> bool:
     # github.com/docker/buildx v0.22.0 Homebrew
     installed_version = stdout.strip().split()[1]
 
-    return installed_version == expected_version
+    if installed_version == expected_version:
+        return True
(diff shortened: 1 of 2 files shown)


devenv/lib/docker.py
```diff
@@ -76,30 +80,78 @@ def _install(url: str, sha256: str, into: str) -> None:
...
+    installed_version = stdout.strip().split()[1]
+
+    return installed_version == expected_version
+
+
 def install_global() -> None:
     version = "27.3.1"
     cfg = {
-        "darwin_x86_64": "https://download.docker.com/mac/static/stable/x86_64/docker-27.3.1.tgz",
+        "darwin_x86_64": f"https://download.docker.com/mac/static/stable/x86_64/docker-{version}.tgz",
```
hubertdeng123, line 109: nit: prob more clear to rename this to version_docker

hubertdeng123 (approved):

PR #189, opened 2025-04-17
feat(doctor): stop docker desktop if it's running

joshuarli:
https://linear.app/getsentry/issue/DI-477/devenv-doctor-should-detect-if-docker-desktop-is-running-and-recommend

### DIFF
--- a/devenv/checks/dockerDesktop.py
+++ b/devenv/checks/dockerDesktop.py
@@ -39,7 +39,7 @@ def fix() -> tuple[bool, str]:
 
     # osascript doesn't wait to make sure it finishes quitting
     # so let's block for up to 5 secs
-    for i in range(10):
+    for _ in range(10):
         time.sleep(0.5)
         if not docker_desktop_is_running():
             return True, ""
--- a/devenv/checks/dockerDesktop.py
+++ b/devenv/checks/dockerDesktop.py
@@ -41,7 +41,7 @@ def fix() -> tuple[bool, str]:
     # so let's block for up to 5 secs
     for i in range(10):
         time.sleep(0.5)
-        if docker_desktop_is_running():
+        if not docker_desktop_is_running():
             return True, ""
 
-    return True, ""
+    return False, "Docker Desktop is taking too long to quit... try again?"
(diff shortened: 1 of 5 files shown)


devenv/doctor.py
```diff
@@ -229,19 +229,19 @@ def main(context: Context, argv: Sequence[str] | None = None) -> int:
         if args.check_only:
             return 1
 
-    print("\nThe following problems have been identified:")
     skip: list[Check] = []
+    print("\nLet's go through the failures one by one.")
     for check in failing_checks:
-        print(f"\t❌ {check.name}".expandtabs(4))
+        print(f"❌ {check.name}".expandtabs(4))
```
asottile-sentry, line 235: the expandtabs aren't doing anything any more

joshuarli: im just gonna remove all of em it hurts my eyes

devenv/checks/dockerDesktop.py
```diff
@@ -0,0 +1,47 @@
...
+        return False, f"failed to quit Docker Desktop:\n{e}\n"
+
+    # osascript doesn't wait to make sure it finishes quitting
+    # so let's block for up to 5 secs
+    for i in range(10):
+        time.sleep(0.5)
+        if docker_desktop_is_running():
+            return True, ""
+
+    return True, ""
```
asottile-sentry, line 47: wouldn't the last one be `False` if it failed to stop it after waiting?

joshuarli: eh i guess it doesn't matter either way, it's just one less docker_desktop_is_running call (from the 2nd followup check being called) if this returns True

i'll change it to False to avoid confusion

joshuarli: well acutally I do want to reuse the error message from check so returning True is preferred

asottile-sentry: oh then isn't the loop backwards -- it's always going to return the first try if docker is still running

joshuarli: Oh... you're right. I'm embarrassed lol, I guess my computer's just too fast

asottile-sentry (approved):

PR #191, opened 2025-04-30
fix: remove archive.download dest before downloading if it's a symlink

joshuarli:
https://linear.app/getsentry/issue/DI-748/devenv-docker-buildx-bugged-installs-due-to-docker-desktop-overwriting

### DIFF
--- a/tests/lib/test_archive.py
+++ b/tests/lib/test_archive.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import io
+import os
 import pathlib
 import tarfile
 import time
@@ -169,6 +170,12 @@ def test_download(tmp_path: pathlib.Path, mock_sleep: mock.MagicMock) -> None:
     )
 
     dest = f"{tmp_path}/a"
+
+    # just exercising the delete symlink if exists path here than copypaste this entire test
+    with open(f"{tmp_path}/hi", "wb"):
+        pass
+    os.symlink(f"{tmp_path}/hi", dest)
+
     with mock.patch.object(
         urllib.request,
         "urlopen",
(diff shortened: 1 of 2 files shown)


tests/lib/test_archive.py
```diff
@@ -169,6 +170,12 @@ def test_download(tmp_path: pathlib.Path, mock_sleep: mock.MagicMock) -> None:
     )
 
     dest = f"{tmp_path}/a"
+
+    # just exercising the delete symlink if exists path here than copypaste this entire test
+    with open(f"{tmp_path}/hi", "wb"):
+        pass
```
asottile-sentry, line 176: what does it do for a dead symlink

joshuarli: a dead symlink'll still be removed since os.path.islink is still True

(but os.path.exists will be False which is why it goes after)

asottile-sentry: more specifically this should be a test too I think

joshuarli: done!

tests/lib/test_archive.py
```diff
@@ -203,6 +210,22 @@ def test_download(tmp_path: pathlib.Path, mock_sleep: mock.MagicMock) -> None:
...
+        autospec=True,
+        side_effect=(io.BytesIO(data),),
+    ):
+        archive.download("https://example.com/foo", data_sha256, dest)
+
+    with open(dest, "rb") as f:
+        assert f.read() == data
+
+    dest = f"{tmp_path}/c"
     with pytest.raises(RuntimeError) as excinfo:
```
asottile-sentry, line 229: these shoudl be a bunch of separate test functions not one big one

asottile-sentry (approved):

PR #195, opened 2025-06-25
fix: remove all docker context on colima start/restart

mchen-sentry:
Delete all docker contexts to ensure only colima context gets recreated. This change deprecates support for OrbStack users.

[Linear](https://linear.app/getsentry/issue/DI-704/devenv-delete-all-docker-contexts-during-colima-startrestart)


joshuarli:
Umm, I also have to fix the sentry integration test which is happening since i removed yarn from sentry lol

PR #197, opened 2025-07-15
ref: venv sync should accomodate uv

joshuarli:
https://linear.app/getsentry/issue/DI-1007/allow-uv-with-devenv


joshuarli:
actually i'm just going to recommend that people directly call `uv sync` in their sync.py: https://github.com/getsentry/devenv/pull/198/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R175

uv has implemented everything devenv.lib.venv supports (venv recreation on version mismatch), so let's just keep devenv.lib.venv as is for pip venvs only

PR #200, opened 2025-08-12
feat: add check/fix for docker currentContext is colima

mchen-sentry:
`dockerContext` within the config should always be specified to `colima`. Adding a check/fix.


joshuarli (approved):
thanks!

PR #204, opened 2025-09-15
feat: install pnpm

keeakita:
This will allow devenv to install and manage `pnpm` on its own instead of relying on repo-specific `sync.py` like is currently done in getsentry/sentry. This will allow us to re-use the logic for getsentry-firetower.

## Test plan

Built with tox, run against a properly configured repo requesting the installation of pnpm (PR to follow)

### DIFF
--- a/devenv/lib/node.py
+++ b/devenv/lib/node.py
@@ -9,7 +9,7 @@ from devenv.lib import fs
 from devenv.lib import proc
 
 
-_shims = ("node", "npm", "npx", "yarn", "pnpm")
+_shims = ("node", "npm", "npx", "yarn")
 
 
 def _install(url: str, sha256: str, into: str) -> None:


devenv/lib/pnpm.py
```diff
@@ -0,0 +1,55 @@
...
+        pass
+
+
+def _version(binpath: str) -> str:
+    stdout = proc.run((binpath, "--version"), stdout=True)
+    # 10.16.1
+    return stdout.strip()
+
+
+def install(version: str, url: str, sha256: str, reporoot: str) -> None:
```
joshuarli, line 39: what'll be passed as `url` to here? afaict there aren't any pnpm binary releases

in sentry devenv/sync.py i'm doing it this way:
(btw i have a ticket to implement the same in devenv, just haven't gotten around to it yet: https://linear.app/getsentry/issue/DI-1012/devenv-install-pnpm)

```
def install_pnpm(version: str, reporoot: str) -> None:
    binroot = fs.ensure_binroot(reporoot)

    if installed_pnpm(version, binroot):
        return

    print(f"installing pnpm {version}...")

    # {binroot}/npm is a devenv-managed shim, so
    # this install -g ends up putting pnpm into
    # .devenv/bin/node-env/bin/pnpm which is pointed
    # to by the {binroot}/pnpm shim
    proc.run((f"{binroot}/npm", "install", "-g", f"pnpm@{version}"), stdout=True)

    fs.write_script(
        f"{binroot}/pnpm",
        """#!/bin/sh
export PATH={binroot}/node-env/bin:"${{PATH}}"
exec {binroot}/node-env/bin/pnpm "$@"
""",
        shell_escape={"binroot": binroot},
    )
```

joshuarli:
superseded https://github.com/getsentry/devenv/pull/210

PR #220, opened 2026-01-27
feat: better linux support 

joshuarli:
my way to implement the ideas in https://github.com/getsentry/devenv/pull/216

- `colima`-tagged builtin checks will not run on non-darwin hosts, making `devenv doctor` linux-friendly
- colima et al are now only installed on darwin
- bootstrap exits if docker[d] isn't installed on linux, requiring it
- `devenv fetch sentry` errors if there are any missing required apt packages
  - edit: this is implemented in sentry `devenv/post_fetch.py` so devenv doesn't have to be updated
    - https://github.com/getsentry/sentry/pull/107462


.github/workflows/bootstrap.yml
```diff
@@ -19,7 +19,7 @@ env:
 
 jobs:
   bootstrap:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
mchen-sentry: might want to put this version bump on a separate PR

README.md
```diff
@@ -85,7 +85,7 @@ Everything devenv needs is in `~/.local/share/sentry-devenv`.
     - `direnv`
```
mchen-sentry: I wonder if we should add some more context to the readme about how `post_fetch` can be used for platform dependency installations.

mchen-sentry (approved):
LGTM

billyvg:
@joshuarli I've been trying to get this to run on linux and hitting this issue now:


```
site-packages/devenv/main.py", line 92, in devenv
    return action(context, remainder)
  File "/usr/local/lib/python3.13/site-packages/devenv/colima.py", line 25, in main
    status = colima.start()
  File "/usr/local/lib/python3.13/site-packages/devenv/lib/colima.py", line 149, in start
    status = check()
  File "/usr/local/lib/python3.13/site-packages/devenv/lib/colima.py", line 104, in check
    macos_major_version = int(macos_version.split(".")[0])
ValueError: invalid literal for int() with base 10: ''
```

I then run with CI which will skip this check but then it hits this block of code and fails because `colima` is not installed:

https://github.com/getsentry/devenv/blob/main/devenv/lib/colima.py#L115-L131

